PR #6, opened 2017-06-07
Feature/fix morph animation

AurL:
A fix for moprh target parsing (on primitives instead of meshes)
I also added a method to disable automatic image flipping. 
Does it make sense for you ?

(Also, I wonder if [this](https://github.com/syoyo/tinygltf/blob/devel/tiny_gltf_loader.h#L930) is a leftover ? I thought it was from my PR at the beginning)


AurL:
This PR is wrong, let's close it.

PR #7, opened 2017-06-19
Feature/serialize gltf

AurL:
Hi, 

Here is the current state of the serialization code.
It currently writes a `.gltf` with it's `.bin` in the same directory, with the same base name.

Please note that the following features are not supported yet:
* embedded images
* binary glTF output
* embedded buffers
* multi buffers
(and I probably missed some of them)

The code is not ready to be merged yet but I'm opening the PR in case you want to do some tests.
Feel free to share your comments on it

(btw, would it be possible to add a few labels so that contributors can qualify and tell a bit about the state of the PRs ?

### DIFF
--- a/tiny_gltf_loader.h
+++ b/tiny_gltf_loader.h
@@ -403,7 +403,7 @@ struct BufferView{
   BufferView()
     : byteOffset(0)
     , byteStride(4)
-  {}  
+  {}
 
 };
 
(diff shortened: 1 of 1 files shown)


syoyo:
I have merged this PR into `devel` branch.

I will refactor code and write some tests for serialization.

> (btw, would it be possible to add a few labels so that contributors can qualify and tell a bit about the state of the PRs ?

That would be nice! But I don't know how to add custom labels on github(e.g. `gltf20` label). Will investigate how to add custom labels.

syoyo:
I will close this branch.

I have renamed filename/class in `devel` branch.  Please open another PR against `devel` if you have further modification.

PR #8, opened 2017-06-22
Fix file/define/class name referenced in glview

soma-arc:
File, define, and class names have been changed.
Fix them referenced in examples/glview/glview.cc


syoyo:
Thanks!

PR #19, opened 2017-08-25
fix(row926): a bug that filename is empty when parse glb

jianghaosen:
when parse glb, filename maybe a empty value, it will be wrong


jianghaosen:
why CI build is wrong?

syoyo:
Cool! Merged.

> CI build is wrong?

syoyo:
I have fixed CI build in the recent `devel`

PR #20, opened 2017-08-29
fix(row1728): accessor->normalize is a local variables and forget ini…

jianghaosen:
accessor->normalize is a local variable and forget initial it, so the value may be true or false. Besides, the value of parse glTF do not assign to normalize


syoyo:
Thanks! Merged

PR #21, opened 2017-09-22
Disable wordexp for Android

thestr4ng3r:
Android does not have wordexp. I just tested tinygltf with this fix on Android and it does work.


syoyo:
Thanks!

PR #23, opened 2017-11-06
Parse KHR_Lights_cmn extension

e-schrade:
Parse light sources as defined in the KHR_lights_cmn extension (used by the blender exporter).


syoyo:
Super cool! Thanks!

PR #25, opened 2017-12-07
Fixed typo ("RPEAT")

cemdervis:


syoyo:
Thanks! I'll check failure as soon as possible

cemdervis:
You're welcome.

syoyo:
> Only the premake script seems to crash (invalid "c++11" value)

cemdervis:
Alright, got it.

PR #30, opened 2018-01-08
Simpler material data retreival

Ybalrid:
Hi,

I'm proposing to add some little methods inside the `Parameter` struct.

These doesn't change the semantics of the type in any way, and just provide a simpler way to get the index of a texture map, a scalar value or a color from the `Parameter`. 

The user is expected to check beforehand the actual declared type of the parameter from the `Materia`l. (in the string, form the Material object) before attempting to use one of theses.

Also, this does include the C++11 <array> header. This template from the STL represent a fixed length array of a type declared on the stack.

The "`ColorValue`" type is typedef as an array<double, 4>

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -42,6 +42,7 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <array>
 
 namespace tinygltf {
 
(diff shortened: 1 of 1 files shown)


syoyo:
Super cool! Merged!

PR #32, opened 2018-01-10
Silence a MSVC warning about narrowing conversion

Ybalrid:
Hello again with a minor code change proposal!

We actually want to convert a size_t value into an int value in this return statement.

Fix the annoying MSVC warning by actually casting the bytestride to an int (in the accessor bytestride computation code)


syoyo:
Awesome! Thank you!

PR #40, opened 2018-02-27
Pr raytrace example

Ybalrid:
Hi, this is a PR for my WIP branch for the new ray-tracing example using NanoRT and NanoSG.

In it's current state it is able to load all the geometry related data from meshes in the glTF file specified in `config.json` (index, vertex, normal and UV, the last two reordered in face-varying order for software rendering).

This is still WIP, but since it can load, display, and render  meshes, I thought it was appropriate to PR the changes. Next step will be loading material information (the different texture maps, as the coordinates are here), but to do that correctly, I think we need to implement the math for PBR software rendering in NanoRT. 

(This changeset also update both the travis and appveyor configuration to build the raytracing example after building the sample program for tinygltf. A shell script pulls the executable for `premake5` from their github for Linux (and Mac) releases at it wasn't easily installable into the travis VMs.

This is useful to do because the gcc and clang version we are using are limited (and configured for) strict C++11 compatibility, but on the windows side, vs2015 supports a mix of C++11, C++14, and Microsoft non standard extensions.)


Ybalrid:
(One of the build VM just stalled for 10 minutes on Travis https://travis-ci.org/syoyo/tinygltf/jobs/346754214#L889 you can relaunch the job by hand if you want)

syoyo:
Super awesome! Merged!

The reason travis failed(timed out) is it takes longer compilation times(`glew.c` takes too long time when assigned VM has low CPU power(or low memory))

Ybalrid:
Oh,  This makes sense. It generally works, but sometimes it doesn't pass because of this.  That's not really important.

PR #42, opened 2018-03-02
raytrace: add simple code that load texture data

Ybalrid:
This is just copying the data from the model object to a Texture one, and pushing it back into the scene textures


syoyo:
Awesome! Merged!

PR #45, opened 2018-03-13
Allow custom image loaders

Squareys:
Hi @syoyo !

Thank you for creating and maintaining tinygltf!

I have a use case where I would like to use my own image loading (e.g. to support more formats than stb_image) and added `TinyGLTF::SetImageLoader` to be able to specify a callback that will be called instead of `LoadImageData`, which uses stb_image.

I made sure that by default it behaves exactly as before (I used my own tests for that, because I was unsure how to run your test suite), the custom image loader callback is entirely optional. In addition there is now a `TINYGLTF_NO_STB_IMAGE` compile flag which, if defined, will prevent inclusion on stb_image and remove the default image loader callback (therefore removing the dependency on it).

I recommend checking out the diffs of the last two commits independently, as the other three minor cleanup commits create a bit of noise in the diff.

Thanks in advance and keep up the awesome work.
Cheers, Jonathan.


syoyo:
Super awesome! Thank you! Will soon review it!

syoyo:
Thank you! merged!

PR #51, opened 2018-03-15
Generic extension support

Selmar:
Two issues remaining:
- The lights data exists twice on export; once outside of the gltf specification and once in the extensions.
- A lot of objects are still missing the extensions property.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -385,6 +385,7 @@ using ColorValue = std::array<double, 4>;
 #endif
 
 typedef std::map<std::string, Parameter> ParameterMap;
+typedef std::map<std::string, ParameterMap> ExtensionMap;
 
 struct AnimationChannel {
   int sampler;              // required
@@ -472,8 +473,8 @@ struct Material {
 
   ParameterMap values;            // PBR metal/roughness workflow
   ParameterMap additionalValues;  // normal/occlusion/emissive values
-  ParameterMap extCommonValues;   // KHR_common_material extension
-  ParameterMap extPBRValues;
+
+  ExtensionMap extensions;
   Value extras;
 };
 
@@ -556,7 +557,7 @@ struct PerspectiveCamera {
         ,
         znear(0.0f) {}
 
-  ParameterMap extensions;
+  ExtensionMap extensions;
   Value extras;
 };
 
@@ -568,7 +569,7 @@ struct OrthographicCamera {
 
   OrthographicCamera() : xmag(0.0f), ymag(0.0f), zfar(0.0f), znear(0.0f) {}
 
-  ParameterMap extensions;
+  ExtensionMap extensions;
   Value extras;
 };
 
@@ -581,7 +582,7 @@ struct Camera {
 
   Camera() {}
 
-  ParameterMap extensions;
+  ExtensionMap extensions;
   Value extras;
 };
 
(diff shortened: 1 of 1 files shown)


syoyo:
Thank you for PR. Will review soon.

> The lights data exists twice on export; once outside of the gltf specification and once in the extensions.

Do you have minimal reproducing test data for this?

Selmar:
It can't actually load the lights extension correctly; it's good you asked me to. The nesting concerns I had earlier were valid concerns.

I think we should use a `map<string, Value>` instead of a `map<string, ParameterMap>`.

Selmar:
Here is my test case, using the cube example. `modeltest.gltf` is the imported file, `lightsTest.gltf` is the re-serialized file.

[lightsTest.zip](https://github.com/syoyo/tinygltf/files/1818936/lightsTest.zip)

syoyo:
I have add some modifications and now lights are serialized under `KHR_lights_cmn` as in `modeltest.gltf` in this branch : https://github.com/syoyo/tinygltf/tree/Selmar-generic_extension_support

````
  "extensions": {
    "KHR_lights_cmn": {
      "lights": [
        {
          "color": [
            0,
            1,
            2,
            3
          ],
          "name": "testLight",
          "type": "fake"
        }
      ]
    }
```

Selmar:
Alright. I suppose you want to change it to using a Value instead of ParameterMap before you merge it? And perhaps some input from people who use the existing extensions objects.

syoyo:
> I suppose you want to change it to using a Value instead of ParameterMap before you merge it? 

Yes! you are right.

tiny_gltf.h
```diff
@@ -1704,7 +1705,7 @@ static bool ParseExtensionsProperty(ExtensionMap *ret, std::string* err, const j
     ParameterMap extValues;
     for (; itVal != itValEnd; itVal++) {
       Parameter param;
-      if (ParseParameterProperty(&param, err, values_object, itVal.key(), false)) {
+      if (ParseParameterProperty(&param, err, itVal.value(), itVal.key(), false)) {
```
Selmar: I have just found that this change actually broke the extension parsing.

I will be changing the ExtensionMap to use Value objects instead, so this'll be fixed along the way.

Selmar:
- changed: parsing integer types to `Value` with type `INT_TYPE`
  - motivation: we should either merge `int_value` and `number_value` or treat them differently
    - if we keep a separation, I suggest to rename `NUMBER_TYPE` to `DECIMAL_TYPE` for clarity, because  integers are numbers too
  - previously, if a json parsed value was an integer, it was parsed as a double regardless
- parsing all json types except for `discarded`
  - e.g. for extras
  - `json::value_t::discarded` is ignored
  - conversion to `Value` with type `BINARY_TYPE` not done (not sure what the intention was)
- `Value` entries for `object` and `array` types are not added if they are `empty` or `null`
- `ParseExtrasProperty` returns false if the parsed property is `NULL_TYPE`
  - previously always returned true
  - doesn't seem to be used anywhere
- Serialization also tested and seems to be functional
  - there is currently still a difference in code design between parsing and serialization of value objects (see `SerializeValue`, `ParseExtrasProperty` and `ParseJsonAsValue`)
  - serializing `BINARY_TYPE` is not done
- Added parsing of extension and extras to Texture
  - we will use it for .dds files (`MSFT_texture_dds`)

syoyo:
Thanks! I have merged your commits into https://github.com/syoyo/tinygltf/tree/Selmar-generic_extension_support

Will soon `update loader_example.cc` then merge it into `devel`

Selmar:
merged, but not tested

syoyo:
I have fixed `loader_example` and added initial support for printing extensions. It looks working so merged into `devel`.

PR #52, opened 2018-03-16
Allow opting out of external image loading

Selmar:
We do not require image loading for external files, we have separate texture directories at the moment.

I've also added saving the image->uri when loading a binary gltf. I assumed this was a mistake as well.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -1670,7 +1670,12 @@ static bool ParseImage(Image *image, std::string *err,
     if (IsDataURI(uri)) {
       loaded = DecodeDataURI(&img, uri, 0, false);
     } else {
-      // Assume external .bin file.
+      // Assume external file
+      // Keep texture path (for textures that cannot be decoded)
+      image->uri = uri;
+#ifdef TINYGLTF_NO_EXTERNAL_IMAGE
+      return true;
+#endif
       loaded = LoadExternalFile(&img, err, uri, basedir, 0, false);
     }
 
@@ -1717,10 +1722,11 @@ static bool ParseImage(Image *image, std::string *err,
       }
     } else {
       // Assume external file
-
       // Keep texture path (for textures that cannot be decoded)
       image->uri = uri;
-
+#ifdef TINYGLTF_NO_EXTERNAL_IMAGE
+      return true;
+#endif
       if (!LoadExternalFile(&img, err, uri, basedir, 0, false)) {
         if (err) {
           (*err) += "Failed to load external 'uri' for image parameter\n";


syoyo:
Cool! Merged!

We are better to introduce option flags to control parsing behavior in the future, not by defining macros.

Selmar:
I guess I agree. I was following the other implementation. I'll keep it in mind if it comes up again!

tiny_gltf.h
```diff
@@ -1685,7 +1685,12 @@ static bool ParseImage(Image *image, std::string *err, const json &o,
     if (IsDataURI(uri)) {
       loaded = DecodeDataURI(&img, uri, 0, false);
     } else {
-      // Assume external .bin file.
+      // Assume external file
+      // Keep texture path (for textures that cannot be decoded)
+      image->uri = uri;
+#ifdef TINYGLTF_NO_EXTERNAL_IMAGE
+      return true;
```
Squareys: @syoyo @Selmar This broke embedded image loading, because this function would ususally not return here, but fail the following LoadExternalFile call after, then load as embedded file and *assign bufferView*, which it didn't up to this point.

How about this instead:

~~~
#ifndef TINYGLTF_NO_EXTERNAL_IMAGE
      loaded = LoadExternalFile(&img, err, uri, basedir, 0, false);
#endif
~~~

Selmar: I see, it goes into that loop for embedded images, even though a bufferview is defined instead of an uri. Perhaps it makes more sense to first check if a bufferview is defined? Otherwise it would try to load from binary data if there is a normal image uri.

Squareys: This basically means that external URIs are always preferred over embedded images. So if both are defined it will use the external. (Doesn't make too much sense imho, but that's a different problem).
I would suggest to go with using the macro to skip the LoadExternalFile call for now, as fixing that also will probably require refactoring the entire function (Won't stop you, though, if you insist ;) ).

syoyo: How glTF 2.0 spec says about this situation? > if both are defined it will use the external.

Selmar: The spec says only one of them should be defined:
>Each image contains one of
>
>  - a URI to an external file in one of the supported images formats, or
>  -  a URI with embedded base64-encoded data, or
>  -  a reference to a bufferView; in that case mimeType must be defined.


The `image.bufferView` reference also says:
> The index of the bufferView that contains the image. Use this instead of the image's uri property.

So if anything, I think we should prefer `bufferView` over `uri` if both are defined.

syoyo: Thanks! So I think I need to add `isEmbedded`(Use bufferView)  flag to an appropriate place in this `if` logic: https://github.com/syoyo/tinygltf/blob/devel/tiny_gltf.h#L1682

Squareys: isEmbedded is already determined a couple of lines above (and uses bufferView.)

syoyo: I have checked schema file

https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/image.schema.json

```
    "oneOf": [
        { "required": [ "uri" ] },
        { "required": [ "bufferView" ] }
    ]
```

So  `bufferView` and `uri` cannot be co-exist(results in invalid glTF data) >  I think we should prefer bufferView over uri if both are defined.

syoyo: So at least we need to report an error when both `uri` and `bufferView` are defined.

syoyo: https://github.com/syoyo/tinygltf/commit/8e2c24fb923434a3dae84a773946f9bee8193199

I did some code refactoring and added more error check in this branch so that the modified code satisfies glTF 2.0 schema.  @Selmar @Squareys Can you please review this changes?

Squareys: Sure, I added three comments to that commit. Nothing dramatic, just minor suggestions.

syoyo: Thanks! I did more code clean-ups based on your review : https://github.com/syoyo/tinygltf/commit/246654a15a60c091c99847ed6c2652ca01fee369

Squareys: Looks great, thanks! 👍

syoyo: Thanks! Merged to `devel`!

nickenchev:
This is great! Was just about to file an issue but got sorted out. Only thing I'd note is that I included the following defines to fully remove image loading.

```
#define TINYGLTF_NO_EXTERNAL_IMAGE
#define TINYGLTF_NO_STB_IMAGE
#define TINYGLTF_NO_STB_IMAGE_WRITE
```

I'm using libPNG for all internal image loading so I didn't want to have to include the ```stb_image``` stuff at all.

Thanks!

PR #53, opened 2018-03-19
Gltf utility

Ybalrid:
Hi,

I'm going to get back to you on this, but here's the basis for a small command line utility for manipulating gltf files called "gltfutil". This is just preliminary work.

Like the validator and the other ones, it's located inside the "example" directory and uses tinygltf.

The only feature it has right now is dumping all the textures of a file as images from a png files (for now, it's using stb_image_write, so there's support for tga and bmp "for free" too, I have a config variable for that, but you can't change it yet).

There's the basic skeleton for taking command lines argument, interpreting them, then executing the asked command.

If you are interested, I would like to know the kind of features that could be added to this, we quickly discussed this some time ago, here's a list of things that would be nice to have: 

  - resizing/recompressing images to make files smaller (especially for binary glb files)
  - manipulating the content of gltf files
    - adding removing meshes
    - changing default scene
    - packing/unpacking multiples scenes into the same glTF
 - converting gltf + bin/images to a single glb file (for this, tinygltf should be able to write back a binary, I see comments about this in the library, but it looks like it's not implemented yet)

What are your thoughts on this?

### DIFF
--- a/examples/gltfutil/CMakeLists.txt
+++ b/examples/gltfutil/CMakeLists.txt
@@ -1,11 +1,9 @@
-cmake_minimum_required (VERSION 3.6)
-project (gltfutil)
+cmake_minimum_required(VERSION 3.6)
+project(gltfutil)
 
 set(CMAKE_CXX_STANDARD 11)
 
-include_directories(
-    ../../
-    )
+include_directories(../../)
 
 file(GLOB gltfutil_sources *.cc *.h)
 add_executable(gltfutil ${gltfutil_sources})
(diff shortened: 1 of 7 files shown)


syoyo:
Cool! will soon revew PR!

Ybalrid:
I opened a PR so you can comment on this. as I said, I've just started working on this. 

Image data is written to disk via the std_image_write library. I plan to use stb_image_resize for adding the option to make the images smaller. 

For re-writing back as GLB, the `bool TinyGLTF::WriteGltfSceneToFile()`in tinygltf will need to be extended. I looked a bit on the layout of these binary files, it looks like it will not be too hard to do.

syoyo:
Cool! Merged!

syoyo:
> resizing/recompressing images to make files smaller (especially for binary glb files)

This feature would be better to have. And as you suggested, Using `stb resize` is the best way to resize image in C++.

PR #54, opened 2018-03-19
Fix MSVC build: removing static keyowrd befor tinygltf::LoadImageData

Ybalrid:
This is peculiar:
It looks like MSVC doesn't want to build when multiple compilation units
(e.g. multiple .cc files) uses tiny_gltf and this funciton is static.

The only one file that will build without error is the one that defined
the `TINYGLTF_IMPLEMENTATION` macro.

Removing `static` here solves the issue...


syoyo:
Good catch! Merged!

PR #58, opened 2018-03-28
Serializer skips null values

bowald:
Serializer skips uninitialized values.

All values that are skipped are optional according to gltf standard.


syoyo:
Great work! Merged!

PR #59, opened 2018-03-30
Added embed buffers as an option in serializer

bowald:
### Summary
- Extended function WriteGltfSceneToFile with embedBuffers argument, default is **false**.
- Added function base64_encode from base64  lib.
- Saver example also outputs embedded gltf.
  
### Question
Did an unsafe cast from size_t to unsigned int to get rid of compiler warning (line 3392), do we want it this way? I didn't want to edit the original implementation of base64_encode.

### DIFF
--- a/examples/saver/main.cc
+++ b/examples/saver/main.cc
@@ -17,6 +17,7 @@ int main(int argc, char *argv[])
   std::string err;
   std::string input_filename(argv[1]);
   std::string output_filename(argv[2]);
+  std::string embedded_filename = output_filename.substr(0, output_filename.size() - 5) + "-Embedded.gltf";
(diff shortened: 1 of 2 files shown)


syoyo:
Awesome! Merged!

> Did an unsafe cast from size_t to unsigned int to get rid of compiler warning (line 3392)

Need to use `static_cast` for the cast to follow C++ standard. I'll fix it.

syoyo:
Fixed in this commit(and also applied clang-format) : https://github.com/syoyo/tinygltf/commit/719d7e4a74969633bde46624d60907aaf329163f

And `size_t` -> uint32 cast should be safe.

PR #60, opened 2018-04-01
Output image data, both embedded and as files.

bowald:
Hi!
Sorry for the wall of text. If you have any feedback on this PR, don't hesitate. I have a lot of time for this tomorrow as well. 

### Summary
- Added **stb_image_write.h** as dependency.
- User can output tinygltf::Image.image as file.
- User can output tinygltf::Image.image as embedded file.
- User can skip image output with define **TINYGLTF_NO_STB_IMAGE_WRITE**. (Tinygltf has same behavior as before this PR).
- User have option to set custom image writer with function SetImageWriter.
- Uninitialized sampler value in tinygltf::Texture is not outputted.
- Updated saver example.
- Updated main readme to reflect the changes.

### Image writing
WriteImageData is a new function which is called for each tinygltf::Image during Serialization. This function can either handle image writing or image embedding, the input tinygltf::Image.uri will be updated with the new data. The updated tinygltf::Image is then serialized (Same as before this PR).  The output image format is the same as the input image format.

If  **TINYGLTF_NO_STB_IMAGE_WRITE** is defined, WriteImageData function will be initialized to nullptr, the user can either set a custom WriteImageData function or skip it entirely. If WriteImageData is nullptr, the tinygltf::Image object is not updated and tinglyf will behave as before the PR. If a custom WriteImageData is set, the user have to update the tinygltf::Image.uri with the new data (either relative path or embedded data). A user may want to have a custom WriteImageData to use a more optimized image writing framework other then stb_image_write.h . 

#### Error handling
As of now, error handling is skipped during serialization. Do we want to have an error string which can return errors to the user, similar to tinyGLTF::LoadASCIIFromFile? As of now I have left comments in the code where an error could happen. 

#### DecodeDataURI change
I wanted to keep the image format from the original file. Therefore the DecodeDataURI function was changed to also set the mime_type. For buffers this is not needed, and the mime_type is discarded (see line 2037). Should this get mime_type be refactored to an own function or is should we keep it as in this PR?

#### Gif format
stb_image_write.h don't support the .gif format. Do we want to support gifs as output? If so, is there a better header only lib other than stb_image_write or should multiple image writing libraries be used?

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,11 +73,12 @@ TinyGLTF uses the following third party libraries.
 * json.hpp : Copyright (c) 2013-2017 Niels Lohmann. MIT license.
 * base64 : Copyright (C) 2004-2008 René Nyffenegger
 * stb_image.h : v2.08 - public domain image loader - http://nothings.org/stb_image.h
+* stb_image_write.h : v1.09 - public domain image writer - http://nothings.org/stb_image_write.h
 
 
 ## Build and example
 
-Copy `stb_image.h`, `json.hpp` and `tiny_gltf.h` to your project.
+Copy `stb_image.h`, `stb_image_write.h`, `json.hpp` and `tiny_gltf.h` to your project.
 
 ### Loading glTF 2.0 model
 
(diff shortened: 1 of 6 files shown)


syoyo:
Super awesome!

I'll review code soon, but firstly Travis failed to the following error.

```
./tiny_gltf.h:2037:12: error: no matching function for call to 'DecodeDataURI'
      if (!DecodeDataURI(&buffer->data, std::string(), buffer->uri, bytes,
           ^~~~~~~~~~~~~
./tiny_gltf.h:1479:13: note: candidate function not viable: expects an l-value for 2nd argument
static bool DecodeDataURI(std::vector<unsigned char> *out,
```

syoyo:
I have no strong motivation to support GIF output, but you can find some GIF writer libraries here: https://github.com/nothings/single_file_libs

Ybalrid:
If I can give my opinion too, I don't think GIF is an useful format for texture support...

For the Travis failing, I think you meant to type `buffer->data()` in that function call. :wink;

(You are currently giving it a function pointer, not the result of the call to `vector<>::data()`)

syoyo:
We can skip error handling for a while. At some point, I'd also want to consider more expressive error report both for loader and writer(In this phase, it would be better to move to schema based approach for generating loader and writer code: https://github.com/syoyo/tinygltf/issues/10)

Ybalrid:
If I can give my opinion, I don't think there's a good reason to support Gif as a texture format either.

bowald:
I agree with you guys, the only reason I mention .gif is due to it being supported by the image loader.  If there is a demand for it, I will be happy to add support for it.

syoyo:
> I mention .gif is due to it being supported by the image loader.

I see. If it is not so hard to add GIF writing feature(e.g. just few additional lines of coding is enough), yes please add  support for GIF writer.

bowald:
> If it is not so hard to add GIF writing feature(e.g. just few additional lines of coding is enough)

stb_image.h supports .gif (used as image loader)
stb_image_write.h do not. (used as image writer)

I could add support by adding another dependency. Such as [gif.h](https://github.com/ginsweater/gif-h), but if it is not needed I think it is better to limit the number of dependencies.

syoyo:
Merged with some modification(suppress clang warnings).

> I could add support by adding another dependency. Such as gif.h, but if it is not needed I think it is better to limit the number of dependencies.

Okay, we can leave GIF support for a while.

PR #64, opened 2018-04-09
Serialization fixes

bowald:
### Summury
- Skips identity matrices for nodes during serialization.
- Skips empty ExtensionMap for materials during serialization
- Null values in ExtensionMap are serialized as empty object **{}** instead of being skipped (An extension witch contain null as value is [KHR_materials_unlit](https://github.com/KhronosGroup/glTF/pull/1163) )
- SerializeGltfParameter outputs Factor if numberArrays size is 1.
- Clang format tiny_gltf.h 

To easier see my changes, see commit 6e96b12a503e47843a7c8aa9f595bb5f3290ff3f, first commit is only clang format

### Discussion
Should default values (such as identity matrices) be skipped during serialization? Or is it up to the user of tinygltf to input correct data?

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -3487,7 +3487,16 @@ bool TinyGLTF::LoadBinaryFromFile(Model *model, std::string *err,
 // GLTF Serialization
 ///////////////////////
 
-// typedef std::pair<std::string, json> json_object_pair;
+static bool isIdentityMatrix(const std::vector<double> &matrix) {
+  if (matrix.size() != 16) return false;
+
+  return matrix[0] == 1.0 && matrix[1] == 0.0 && matrix[2] == 0.0 &&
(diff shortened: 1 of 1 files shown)


syoyo:
> Should default values (such as identity matrices) be skipped during serialization? 

Yes, we can safely skip serializing default values where JSON schema defines it : https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/node.schema.json

bowald:
> Yes, we can safely skip serializing default values where JSON schema defines it

Good, I will keep adding checks against default values during serialization.

syoyo:
@bowald Any updates?

bowald:
@syoyo I'm sorry, been busy. will have some time to make fixes in the near coming weeks.

PR #69, opened 2018-04-24
Serialization fixes for single-number values and integers

benbuzbee:
- When serializing single number property values, such as the pbrMetallicRoughness properties like metallicFactor, do not put them in a size 1 array. This fixes validation against the [Khronos group GLTF validator
](http://github.khronos.org/glTF-Validator/)

- When outputting texture indices, do so as an integer, not a floating point. This extra strictness allows some importers, such as Microsoft's, to import the GLTF file

- When outputting integer arrays, do not make them floating points. This extra strictness allows some importers, such as Microsoft's, to import the GLTF file

- Re-apply clang formatting based on the .clang-format file

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -42,7 +42,6 @@
 #include <cstdint>
 #include <cstring>
 #include <map>
-#include <optional>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


syoyo:
Cool!

> The CI build appears to run with a version of the STL without std::optional

`std::optional` is C++17 feature. At the moment TinyGLTF is targeting for C++11 compilers. Can you rewrite it so that code can be compiled with a C++11 compiler?

syoyo:
Awesome! Merged!

PR #73, opened 2018-05-24
Added missing extras serialization for nodes

holzon:
Noticed that extras field was not serialized for nodes (and many other elements).
(The existing extras serialization for scene and asset seem to check for the existance of keys in the element, but according to the spec, extras can be any type, not only objects.)


syoyo:
Thanks! Merged!

PR #75, opened 2018-06-04
Adding missing extras field serializaton

holzon:
Added serialization of the rest of the missing extras fields from the gltf 2.0 specification.


syoyo:
Super cool! Merged!

PR #80, opened 2018-07-24
Import texture names as well

mosra:
Every other structure has it already, so it's probably just an overlook. Simply copied the parser line from other places. To be extra sure, I double checked the spec that textures really can have the `"name"` field as well.

Thank you! :+1:


syoyo:
Awesome! merged.

PR #81, opened 2018-07-24
Don't try to expand file paths on Emscripten.

mosra:
Since the same is done on all mobile platforms already, I think it should not do any harm. Without this I was getting compilation/linker errors.


syoyo:
Thanks! merged.

PR #86, opened 2018-08-16
Properly fail also when external *.bin are missing from *.glb.

mosra:
When `*.bin` files referenced from `*.gltf` files are not found, the loading fails (as it should), but for `*.glb` it silently passed. Fixed that to make the behavior consistent.


syoyo:
Thanks! merged.

PR #88, opened 2018-08-22
forward declare DataURI helpers

Selmar:
forward declare DataURI helper functions to allow usage outside of implementation file (in case of custom image handling)

if you want anything done differently about it I'll do that of course


syoyo:
Thanks! Merged.

PR #89, opened 2018-08-22
missing bin files are reported as warning instead of error

Selmar:
failing to open is an error, so I believe this should be an error as well

in addition, the warning string is not even used here, so I had to step through the code to see what went wrong

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -1238,7 +1238,7 @@ std::string base64_decode(std::string const &encoded_string) {
 
 static bool LoadExternalFile(std::vector<unsigned char> *out, std::string *err,
                              std::string *warn, const std::string &filename,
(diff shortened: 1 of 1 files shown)


syoyo:
Thank you for PR.

`LoadExternalFile` is also used for loading texture files. For texture file, I think it should be better to report it as an warning so that people can at least load .gltf file without textures. .bin is mandatory thus it should report as an error.

So I think it would be nice to add `must_exist` or `mandatory` argument to `LoadExternalFile` and decide making warning or error for missing file.

Selmar:
`required` perhaps, for consistency?

syoyo:
Yeah, `required` should be nice!

Selmar:
this will do?

syoyo:
Super awesome! Merged!

Issue #56, opened 2018-03-25
Force conversion textures into 32-bit (RGBA) support (Vulkan API, AMD) in native image loader


elviras9t:
Latest AMD drivers does not support RGB format in textures. Only R, RG, or RGBA.

syoyo:
You can send PR

syoyo:
RGBA is now default  with PR#90

PR #90, opened 2018-08-23
Force default image loader to use 32-bit images

victorbush:
Simple fix for #56.

Forces stbi library to output 4 channels when loading images, resulting in 32-bit textures even if the source image is only 24-bit (or less).

It seems that many (?) AMD and NVIDIA devices do not support 24-bit images in Vulkan.

Does forcing 32-bit textures in the default image loader seem reasonable or should this be configurable?

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -1314,14 +1314,19 @@ bool LoadImageData(Image *image, std::string *err, std::string *warn,
                    int size, void *) {
   (void)warn;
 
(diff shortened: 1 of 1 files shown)


syoyo:
Thanks! merged.

> Does forcing 32-bit textures in the default image loader seem reasonable or should this be configurable?

This is todo. I'm planning to add an option for configuring default texture channels.

PR #92, opened 2018-08-28
handling stbi_write_xxx_to_func return code

walf443:
sbti_write_xxx_to_func function return int result. if result is 0, it failed. and result is 1, it successed. So WriteImageData should consider error code.

* [sbti_write_png_to_func](https://github.com/syoyo/tinygltf/blob/master/stb_image_write.h#L1241)
* [sbti_write_jpg_to_func](https://github.com/syoyo/tinygltf/blob/master/stb_image_write.h#L1736)
* [sbti_write_bmp_to_func](https://github.com/syoyo/tinygltf/blob/master/stb_image_write.h#L457)

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -1385,18 +1385,24 @@ bool WriteImageData(const std::string *basepath, const std::string *filename,
   std::vector<unsigned char> data;
 
   if (ext == "png") {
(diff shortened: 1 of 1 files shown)


syoyo:
Awesome! merged.

walf443:
Thanks your fast review!

PR #93, opened 2018-08-30
Enable support for writing binary glTF.

davidharmon:


syoyo:
Awesome! merged!

There is Travis error but we can ignore it for now(they claims clang-3.7 is not found)

davidharmon:
Great!

Issue #94, opened 2018-09-24
Validation for serialization


Selmar:
Hi,

I'm wondering if there is currently any way to validate that serialization is working correctly? The project I'm working on has many and rather large models, so any mismatch is a potentially large time-waster. There have been a few errors before (as I've seen and submitted pull requests to fix them), so this could help in both catching mistakes and testing new implementations. 

E.g. I would like to compare model data using the equals `==` operator. If it is a wanted feature, I could make a pull request.

Alternatively, I imagine the json parser implements the equals operator, so perhaps we could refactor to make use of that?

Any ideas and concerns are welcome.

syoyo:
There is a validator example using json-schema-validator https://github.com/syoyo/tinygltf/tree/master/examples/validator , so you can run it as a postprocess after serialization or in an unit tester.
(Or you could use python script if you don't need to stick with C++11) 

> Alternatively, I imagine the json parser implements the equals operator, so perhaps we could refactor to make use of that?

Do you mean you'd like to compare input  glTF and glTF schema with `==` operator?

It could be possible and rather easy, but in such a case we need to include/prepare glTF schema files in some way(stringify and embed glTF schema files into .h are one possible solution)

Selmar:
In my case I really want to validate the contents of the files actually, e.g. missing parameters that exist in one but not the other, or mismatching values due to forgetting to serialize something. I'm not too worried about the schema, because if the format is not correct that will either result in a crash or show itself otherwise (e.g. through missing parameters or indeed through any other validation methods we use).

syoyo:
> I already implemented the == operator locally

Good! Is this feature only requires current tinygltf dependency(i.e. json.hpp), if so PR is appreciated.

It would also be nice if we could have out-of-bounds check for integer indices(node, texture, target, etc) which cannot be validated using JSON schema.

Selmar:
Do you prefer the `operator==()` declaration and/or implementation inside or outside of the class definition?

1.
```
class ClassName
{
  bool operator==(const ClassName& other); // ... declaration
}
// ... implementation
```

2.
```
class ClassName
{
  bool operator==(const ClassName& other)
  {
    // ... implementation with declaration
  }
}
```

3.
```
class ClassName;
bool operator==(const ClassName& one, const ClassName& other); // ... declaration
...
class ClassName
{
}
...
bool operator==(const ClassName& one, const ClassName& other)
{
  // ... implementation
}
```

syoyo:
[...]

PR #95, opened 2018-10-03
Equality operator and some additions

Selmar:
Changelist:
- equality operator for tinygltf::Model and types contained by it
- double comparison
- initialize some variables for operator==() reliabiltiy
- change some floats to doubles
- intellisense define
- serialize accessor.name
- dont serialize some names if they're empty
- return false if writing gltf file fails

Discussion:
- Double comparison method is perhaps not necessary as the json serializer writes/reads the bytes from/into memory directly. I think my previous issue came from the fact that some variables were saved in memory as `float`, but serialized as `double`. Thoughts?
- Necessary to reformat for consistency, perhaps you want to do that yourself?

Potential issues in current state as mentioned in #94, related to PR:
- Comparing pre-serialization and post-serialization model data will not work for the buffer property and has to be handled separately if the user wants to verify it, because currently the buffer.uri property is coded to copy the gltf file name and replace the extension instead of keeping the previous buffer uri.
- Comparing pre-serialization and post-serialization model data will not work for neither buffers nor images if they are embedded, because for obvious reasons the uri changes. This has to be handled separately by the user.

Potential issues in current state as mentioned in #94, unrelated to PR:
- Looking at the code, loading embedded buffers seems to take up more memory/cpu, which can be significant for large models. The embedded data is completely stored in the buffer.uri property and then duplicated in the buffer.data property, but for external buffers it's only loaded into buffer.data directly.
- Every buffer is saved with the same filename, which should either overwrite the previously written buffer or generate an exception.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -41,6 +41,7 @@
 #include <cassert>
 #include <cstdint>
 #include <cstring>
+#include <cstdlib>
 #include <map>
 #include <string>
 #include <vector>
@@ -133,6 +134,9 @@ namespace tinygltf {
 
 #define TINYGLTF_SHADER_TYPE_VERTEX_SHADER (35633)
 #define TINYGLTF_SHADER_TYPE_FRAGMENT_SHADER (35632)
+
+#define TINYGLTF_DOUBLE_EPS (1.e-12)
+#define TINYGLTF_DOUBLE_EQUAL(a,b) (abs((b)-(a))<TINYGLTF_DOUBLE_EPS)
 
 typedef enum {
   NULL_TYPE = 0,
@@ -293,6 +297,8 @@ class Value {
   }
 
   size_t Size() const { return (IsArray() ? ArrayLen() : Keys().size()); }
+  
+  bool operator==(const tinygltf::Value& other) const;
 
  protected:
   int type_;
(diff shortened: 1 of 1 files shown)


Selmar:
And perhaps unrelated; i see you have merged some more changes into the `devel` branch a while back. Wasn't it obsolete?

syoyo:
Thanks!

> Double comparison method is perhaps not necessary as the json serializer writes/reads the bytes from/into memory directly. I think my previous issue came from the fact that some variables were saved in memory as float, but serialized as double. Thoughts?

clang emit warning when using `==` for floating point comparison. 
Also there may be precision loss when loading/storing fp value from/to memory on different CPU architecture, so using fp comparator is much safer.

> And perhaps unrelated; i see you have merged some more changes into the devel branch a while back. Wasn't it obsolete?

Oh, I've merged this https://github.com/syoyo/tinygltf/commit/098dfee98209aa1a447c7f7f65123f7e53b4d0c3 to `master`

syoyo:
> Necessary to reformat for consistency, perhaps you want to do that yourself?

Suppressed some clang warnings and applied clang-format:

https://github.com/syoyo/tinygltf/commit/641b3ccf8c3549f8f45f8da02cf8ac49a1c1149c

PR #96, opened 2018-10-05
Serialization & comparison fixes

Selmar:
Some more fixes.

- Parse image extras property
- Optionally serialize texture.source (it is not required in the current spec)
- Always check Parameter::number_value to catch user errors, e.g. setting number_value without setting has_number_value, which will cause it to serialize as a bool but read back in as a float anyway
- Support writing multiple bin files
- Use existing buffer.uri as filename if not empty and not a data uri, else a default file name
- Protect against writing the same uri twice (of course naming conflicts when writing multiple gltf+bins to the same directory is up to the user to prevent).

Discussion:
This PR changes the default behavior for writing `.bin` file(s). 
- Previously, only the last buffer in the list would be saved (and `std::ofstream` default is truncate I believe, so the data was really lost). Now all buffers are saved.
- More importantly, the `bin` filename previously defaulted to the glTF filename with a `.bin` extension. Now, the default name is the `buffer.uri` property, if valid, with the previous behavior as a fallback.


syoyo:
Awesome! Merged.

> This PR changes the default behavior for writing .bin file(s).

Current serialization feature is not complete, thus I'd like to adopt to the behavior in your PR.

PR #99, opened 2018-10-08
Pretty print & some (de)serialization fixes

Selmar:
- pretty print as optional parameter, default true
- serialize name instead of type for camera.name
- create empy Value::Object if extension value parsing failed (e.g. empty object), so the type is still object (similar to default behaviour for serialization)


syoyo:
Thanks! merged.

PR #103, opened 2018-10-25
[serialization] handle failing buffer file writes

Selmar:
fail serialization if a buffer file fails to be written (e.g. due to access rights)


syoyo:
Awesome! merged!

PR #105, opened 2018-11-06
Fixed a reference error which cause an error when writing images to disk

bowald:
A reference to the pointer **this->write_image_user_data_** have been passed to **UpdateImageObject()** instead of the actual pointer. 

Also removed trailing whitespaces from two lines.

## Before this PR,
**FsCallbacks** could not be used when serializing image (due to write_image_user_data_ wrongly beeing passed as reference). 

## After PR 
images can be serialized as separate files to disk using **FsCallbacks**.


syoyo:
Thanks! merged.

PR #107, opened 2018-11-09
Add missing serialization of `sampler.wrapR`

Selmar:
Add missing serialization of `sampler.wrapR` .. : )


syoyo:
Awesome! thanks!

PR #110, opened 2018-11-15
Removed unreachable code and fixed ambiguity warning.

AlvaroBarua:
I usually work with the highest level of warnings enabled and these two warnings prevented me from compiling.


syoyo:
Awesome! Thank you!

PR #111, opened 2018-11-15
This is a basic example I added after figuring out how to use the tinygltf library

viperscape:
I created a basic opengl example that uses the modern opengl pipeline. It also renders the base texture. This might be a good starting point to updating the original glview example, I'm not sure that's up to you!

screenshot


syoyo:
Super cool! Merged!

PR #115, opened 2018-12-04
some serialization fixes

Selmar:
- Fix pretty print to really not print pretty (it was printing pretty with indentation 0).
- Default value for `sampler.wrapR`, uninitialized value was causing the `operator==` to fail. It was serialized with an uninitialized value. I remember adding the serialization, but in fact, looking through the code, it never even seems to be deserialized... Merging error? Legacy?


syoyo:
Thanks!

> I remember adding the serialization, but in fact, looking through the code, it never even seems to be deserialized... Merging error? Legacy?

Hmm... I have no idea...

PR #117, opened 2018-12-05
Fill in missing bufferView targets

viperscape:
If the target type is missing, we fill them in. I use auto for loops here, hope that's ok!


syoyo:
Awesome! merged!

PR #119, opened 2018-12-07
Add CMake package for developer

nyue:
Added the glTFConfig.cmake file so that once
installed, it is easy for developers to find
all the path location for include purposes.


syoyo:
Super awesome! Thanks!

PR #122, opened 2018-12-09
Fix incorrect index into bufferViews array

lebek:
From the spec: "When a primitive's `indices` property is defined, it references the accessor to use for index data"

In the spec `accesors.bufferView` is optional, but in tinygltf it's required, so I don't check for the empty case.

@viperscape can you check this still does what you wanted?


viperscape:
Yes that should be better, I missed that accessor step thanks.

PR #125, opened 2018-12-25
Fix bug in handling OS X build via CMake

nyue:
The OS macros needs to be upper case for the
if statement to work as expected


syoyo:
Awesome! Thanks!

PR #127, opened 2018-12-26
Renamed glTFConfig to TinyGLTFConfig

nyue:
This renaming is done so that the config file matches
the project name which is TinyGLTF. This is done to prevent
future name collision should the Khronos glTF project decides
to provide their own CMake configuration file.


syoyo:
Thanks! merged.

PR #129, opened 2018-12-28
Added new compiler option for loading all gltf related files from android app asset package

SaschaWillems:
This PR adds the new compiler option `TINYGLTF_ANDROID_LOAD_FROM_ASSETS` that will load all files, including textures and binaries from the packaged apk instead of the file system using the asset manager.

With this you can package arbitrary gltf files with your Android apk and load them directly from your application package without having to use embedded gltf files.

With this change you can simply take any gltf model like the Polly sample scene, add the directory as an asset to your Android build and load it up like you'd be working with a plain directory.

Demo video: https://youtu.be/rFXgeuyI4H4

P.S.: I'm not 100% sure if the coding style fits your guidelines for this library, so if anything needs to be changed just drop me a line.


syoyo:
Awesome! Thanks!

> P.S.: I'm not 100% sure if the coding style fits your guidelines for this library, so if anything needs to be changed just drop me a line.

Code is formatted with clang-format, thus you don't need to worry about it. I'll apply clang-format then.

Issue #132, opened 2019-01-06
Support 16bit PNG


syoyo:
AntiqueCamera model contains 16bit PNG image: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AntiqueCamera

stb image does not support loading 16bit PNG.

Candidate is use lodepng library: https://lodev.org/lodepng/

Ybalrid:
`lodepng` looks interesting. It just need to be put into a header-only form for it to be used as part of tinygltf

syoyo:
I have alreasdy added singe-header version of lodepng in `lodepng` branch. You may be interested in it: https://github.com/syoyo/tinygltf/tree/lodepng

vinjn:
I think stb_image today supports 16bit png.
https://github.com/nothings/stb/blob/master/stb_image.h#L24
https://github.com/nothings/stb/blob/master/stb_image.h#L61

Ybalrid:
@vinjn Interesting. We should upgrade the stb libraries usded in tinygltf... Our stb_image.h is 3 years old!! 😮 

So. I just upgraded the headers for stb_image(_write).h with the latest version on nothings's repo, and tried to read the texture content of the problematic file (AntiqueCamera). It seems that it reads the texture data via stb_image just fine (using the "texture dumper" of the `gltfutil` sample)

![image](https://user-images.githubusercontent.com/1816251/53683811-734bf780-3d05-11e9-96a9-6b7dd7150049.png)

@syoyo If you don't mind, I'm going to PR the upgraded stb libraries. This is probably a cleaner and simpler solution that integrating lodepng to handle this particular case. 😅

syoyo:
@Ybalrid Awesome!

Yes, please. We can discard lodepng version. > I'm going to PR the upgraded stb libraries

PR #133, opened 2019-01-10
fix ci build of example

Ybalrid:
one of the programs built by the CI scripts wasn't declaring the single-header implementation of lodepng, making the CI build fail for no good reason.

liked to issue #132


Ybalrid:
You know, the ever-important task of making little badges stay green... :smiley:

PR #134, opened 2019-01-24
Initial support for draco mesh compression

abwood:
Initial support for draco mesh compression. In this PR, we establish that draco is a dependency built external to tinygltf, which breaks from the current tradition of header only dependencies. For that reason, this feature is hidden behind a #define TINYGLTF_ENABLE_DRACO and requires developers to explicitly opt-in for draco support.

In this change, tinygltf any primitive using draco compression will:
If indices are specified
1) Decode the index buffer using draco, creating a new buffer and bufferview, adding to Model::bufferViews and Model::buffer collections
2) Update the primitive's accessor id to reference this new decoded bufferview.
For each attribute semantic specified by the draco extension
1) Decode the vertex buffer using draco, creating a new buffer and bufferview, adding to Model::bufferViews and Model::buffer collections
2) Update the primitive's accessor id to reference this new decoded bufferview.

TODO:

- [ ] Update README.
- [ ] Provide unit tests or examples.
- [ ] Consider adding draco as a submodule and update cmake build to build draco

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -538,8 +538,9 @@ struct BufferView {
                       // understood to be tightly packed
   int target;         // ["ARRAY_BUFFER", "ELEMENT_ARRAY_BUFFER"]
   Value extras;
+  bool dracoDecoded;  // Flag indicating this has been draco decoded
 
-  BufferView() : byteOffset(0), byteStride(0) {}
+  BufferView() : byteOffset(0), byteStride(0), dracoDecoded(false) {}
   bool operator==(const BufferView &) const;
 };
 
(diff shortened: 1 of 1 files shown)


abwood:
@syoyo Apologies for the branch name collision.  It appears that we both created draco branches around the same time :smile:

syoyo:
Super awesome! Merged!

> It appears that we both created draco branches around the same time 

Yes, I was also tried to support draco. Your code is much better than me :-)

syoyo:
> Cool. Looks like this got merged into the draco branch. Will it go in master soon?

After confirming loading draco mesh in `examples/glview` example, I'll merge it to `master`.

PR #138, opened 2019-01-30
Add line strip primitive mode define

dolphineye:
Hi,

I'm sending this small pull request to add the `TINYGLTF_MODE_LINE_STRIP` primitive mode which was curiously missing. The constant value of the define was set to `3` as per the [GLTF 2.0 specification](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#primitivemode.).

Best regards!


syoyo:
Awesome! merged!

PR #140, opened 2019-01-31
Adding default values for min and mag filter.

timmmeh:
Hello there, running across a few things as I integrate tinygltf, this one is just a nicety. More to follow :)


syoyo:
Awesome! Merged.

PR #142, opened 2019-02-14
Change AnimationSampler::interpolation to a non-required field.

ebirenbaum:
The spec states to default to LINEAR when this field is not present, so parsing should not fail because a sampler is missing the "interpolation" field.


syoyo:
Awesome! Than you!

PR #144, opened 2019-02-23
Add direct access to texture coordinate set on texture parameter

SaschaWillems:
Similar to the function to easily access the index of a texture (parameter), this function adds easy access for the texture coordinate set of a texture.


syoyo:
Awesome! Merged.

PR #145, opened 2019-03-02
Upgrade the STB libraries, and fix #132

### LINKED ISSUE #132: Support 16bit PNG

Ybalrid:
Thanking @vinjin for pointing out that newer versions of STB image can load 16bit PNGs...

### DIFF
--- a/examples/gltfutil/main.cc
+++ b/examples/gltfutil/main.cc
@@ -20,8 +20,8 @@ int usage(int ret = 0) {
           "[path to output directory])\n\n"
        //<< "\t\t -i: start in interactive mode\n"
        << "\t\t -d: dump enclosed content (image assets)\n"
-       << "\t\t -f: file format for image output"
-       << "\t\t -o: ouptput directory path"
+       << "\t\t -f: file format for image output\n"
+       << "\t\t -o: ouptput directory path\n"
        << "\t\t -h: print this help\n";
   return ret;
 }
--- a/examples/gltfutil/texture_dumper.cc
+++ b/examples/gltfutil/texture_dumper.cc
@@ -1,5 +1,5 @@
-#include <iostream>
 #include <algorithm>
+#include <iostream>
 
 #include "stb_image_write.h"
 #include "texture_dumper.h"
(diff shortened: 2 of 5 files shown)


syoyo:
If you can read AntiqueCamera just uploading STB, it looks recent `stb_load_from_memory` handles 16bit png by downsampling it to 8bit.

It think proper way is to use `stb_load_16_from_memory`(then fallback to `stb_load_from_memeory` if failed) and also record bits and pixel_format into `Image` class as done in `lodepng` branch:
 
https://github.com/syoyo/tinygltf/blob/lodepng/tiny_gltf.h#L481

Ybalrid:
I see what you mean. As is done now we hare throwing away half of the bit depth as each color channel is mapped from 2 bytes to 1 byte.

I think attempting to load the special case, then the general one if it faild is the good aproach indeed. 

Will update the PR later to have this logic implemented in tinygltf.h

Ybalrid:
I think if we call stbi_load_16_from_memroy and that the image *was* 8bit it also convert it to 16 (it does what regular `stbi_load_from_memory` did, but in reverse:

See that line of code from stb_image:
https://github.com/nothings/stb/blob/master/stb_image.h#L1150

So the above stragegy will not work. We need to know *in advance ( = to detect)* that the image is 16bits :  

  1) If the only image format that is 16bits that we are supposed to encounter is 16bits PNGs, maybe reading the first few bytes of the memory buffer ( = the PNG header ) could inform us of the actual file format. 

 2) We could also call their `stbi__load_main` function first to determine the number of bits per channel, technically I think it's STBI's internal API, but we could do that?

Give this information, what way should I pursue to implement this. I think the 2nd is best. Just have to keep in mind that we are doing this when upgrading STBI later down the road, if they change this function behavior (unlikely).

Ybalrid:
This is a good sign: 
image

Ybalrid:
So, example `gltfutil` cannot write the images back as 16bit because `stb_image_write` does not support 16bit. (I've added a little conversion loop there just so we can see if the loaded data looks right. It does. This example program isn't important.)

Now: tinygltf can load 16bit images, with the correct data entred into the Image struct, here's a sample output:

```
image name is: "camera_camera_Normal"
image size is: 2048x2048
pixel channel count :4
pixel bit depth :16
``` 

From an image loaded via tinygltf's default `LoadImageData`

syoyo:
Awesome! Merged!

> example gltfutil cannot write the images back as 16bit because stb_image_write does not support 16bit.

Extending `stb_image_write` to support 16bit may be rather easy. I'll take a look it.

PR #146, opened 2019-03-02
Fix for when json and stb is aready used in project

kroko:
The original code is as follows

```c++
#include "json.hpp"

#ifdef TINYGLTF_ENABLE_DRACO
#include "draco/core/decoder_buffer.h"
#include "draco/compression/decode.h"
#endif

#ifndef TINYGLTF_NO_STB_IMAGE
#include "stb_image.h"
#endif

#ifndef TINYGLTF_NO_STB_IMAGE_WRITE
#include "stb_image_write.h"
#endif
```

First issue is that if project already uses _std_ image stuff, compiler will throw _stb_ related redefinition errors when including _tinygltf_.

Second issue is also related to the fact that _json_ and _stb_ is used, however not by relative paths, but by using header search paths totally elsewher in file system. First issue could be resolved by refactoring project code, but this is a bit different. For example my code has

```c++
#define STB_IMAGE_IMPLEMENTATION
#define STB_IMAGE_WRITE_IMPLEMENTATION
#include <stb/stb_image.h>
#include <stb/stb_image_write.h>
#include <json/json.hpp>
```

Now, obviously _tinygltf_ will fail as there is no header at `#include "header_name.ext"`.

It might not be the most elegant way, but it is very explicit, by introducing 

```c++
TINYGLTF_BYPASS_INCLUDE_JSON
TINYGLTF_BYPASS_INCLUDE_STB_IMAGE
TINYGLTF_BYPASS_INCLUDE_STB_IMAGE_WRITE
```

So now we get the final pull:

```c++
#ifndef TINYGLTF_BYPASS_INCLUDE_JSON
#include "json.hpp"
#endif

#ifdef TINYGLTF_ENABLE_DRACO
#include "draco/core/decoder_buffer.h"
#include "draco/compression/decode.h"
#endif

#ifndef TINYGLTF_NO_STB_IMAGE
#ifndef TINYGLTF_BYPASS_INCLUDE_STB_IMAGE
#include "stb_image.h"
#endif
#endif

#ifndef TINYGLTF_NO_STB_IMAGE_WRITE
#ifndef TINYGLTF_BYPASS_INCLUDE_STB_IMAGE
#include "stb_image_write.h"
#endif
#endif
```

And user code can look something like this

```c++
// In practive this is equivalent to
//
//#define TINYGLTF_IMPLEMENTATION
//#define STB_IMAGE_IMPLEMENTATION
//#define STB_IMAGE_WRITE_IMPLEMENTATION
//#include "tiny_gltf.h"
//
// But here I have the ability to specify paths to json and stb myself
//
#define TINYGLTF_IMPLEMENTATION
#define STB_IMAGE_IMPLEMENTATION
#define STB_IMAGE_WRITE_IMPLEMENTATION
#define TINYGLTF_BYPASS_INCLUDE_JSON
#define TINYGLTF_BYPASS_INCLUDE_STB_IMAGE
#define TINYGLTF_BYPASS_INCLUDE_STB_IMAGE_WRITE
#include <json/json.hpp>
#include <stb/stb_image.h>
#include <stb/stb_image_write.h>
#include <tinygltf/tiny_gltf.h>
```

Or in case where project already uses _stb_ and/or _json_ somewhere up in the tree no refactoring is needed 

```c++
#define STB_IMAGE_IMPLEMENTATION
#define STB_IMAGE_WRITE_IMPLEMENTATION
#include <stb/stb_image.h>
#include <stb/stb_image_write.h>
#include <json/json.hpp>

// ...

#define TINYGLTF_IMPLEMENTATION
#define TINYGLTF_BYPASS_INCLUDE_JSON
#define TINYGLTF_BYPASS_INCLUDE_STB_IMAGE
#define TINYGLTF_BYPASS_INCLUDE_STB_IMAGE_WRITE
#include <tinygltf/tiny_gltf.h>
```

By using these bypass flags _tinygltf_ can use _json_ and _stb_ without introducing clashes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,10 @@ if (!ret) {
 * `TINYGLTF_NO_EXTERNAL_IMAGE` : Do not try to load external image file. This option woulde be helpful if you do not want load image file during glTF parsing.
 * `TINYGLTF_ANDROID_LOAD_FROM_ASSETS`: Load all files from packaged app assets instead of the regular file system. **Note:** You must pass a valid asset manager from your android app to `tinygltf::asset_manager` beforehand.
 * `TINYGLTF_ENABLE_DRACO`: Enable Draco compression. User must provide include path and link correspnding libraries in your project file.
+* `TINYGLTF_BYPASS_INCLUDE_JSON `: Disable including `json.hpp` from within `tiny_gltf.h` (usually because it has been included before).
+* `TINYGLTF_BYPASS_INCLUDE_STB_IMAGE `: Disable including `stb_image.h` from within `tiny_gltf.h` (usually because it has been included before).
+* `TINYGLTF_BYPASS_INCLUDE_STB_IMAGE_WRITE `: Disable including `stb_image_write.h` from within `tiny_gltf.h` (usually because it has been included before).
+
 
 ### Saving gltTF 2.0 model
 * [ ] Buffers.
--- a/README.md
+++ b/README.md
@@ -134,9 +134,9 @@ if (!ret) {
 * `TINYGLTF_NO_EXTERNAL_IMAGE` : Do not try to load external image file. This option woulde be helpful if you do not want load image file during glTF parsing.
 * `TINYGLTF_ANDROID_LOAD_FROM_ASSETS`: Load all files from packaged app assets instead of the regular file system. **Note:** You must pass a valid asset manager from your android app to `tinygltf::asset_manager` beforehand.
 * `TINYGLTF_ENABLE_DRACO`: Enable Draco compression. User must provide include path and link correspnding libraries in your project file.
-* `TINYGLTF_BYPASS_INCLUDE_JSON `: Disable including `json.hpp` from within `tiny_gltf.h` (usually because it has been included before).
-* `TINYGLTF_BYPASS_INCLUDE_STB_IMAGE `: Disable including `stb_image.h` from within `tiny_gltf.h` (usually because it has been included before).
-* `TINYGLTF_BYPASS_INCLUDE_STB_IMAGE_WRITE `: Disable including `stb_image_write.h` from within `tiny_gltf.h` (usually because it has been included before).
+* `TINYGLTF_BYPASS_INCLUDE_JSON `: Disable including `json.hpp` from within `tiny_gltf.h` because it has been already included before or you want to include it using custom path before including `tiny_gltf.h`.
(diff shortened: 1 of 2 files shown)


syoyo:
Aangle-backed include is basically for system header files. We don't recommend to use it for stb/json inclusion.

https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html


> compiler will throw stb related redefinition errors when including tinygltf.

What is actual compiler error log? It should not report redefinition errors if you only define one `***_IMPLEMENTATION` in .cc/c++ files.

kroko:
To quote the the link, I fall into the category who

> (..) specify other options on the command line, such as -I, that affect where the preprocessor searches for header files (..)

and IMHO this

> (..) The most commonly-used option is -I (..)

is true.
I tend to specify header search paths for external project dependencies using `-I` and I see nothing wrong with it. Of course I could change habits, start using [`-iquote`](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html) or other. 

Anyways, don't you think that ability to override default hardcoded

```cpp
#include "json.hpp"
#include "stb_image.h"
#include "stb_image_write.h"
```

in `tiny_gltf.h` is a nice option to have?

Plus IMHO IDEs may use `-I` automatically under the hood, when in shiny GUI one specifies search paths (thus not explicitly dragging those dependencies into the project tree) (for example Xcode).

If you agree, but don't like naming of the the definitions (as said - very explicit) I can change them as you like! 😄 

And pardon - forget about redefinition errors, it was caused in conjunction with me trying to solve hardcoded "quoted" includes without refactoring existing include tree hierarchy of the project.

syoyo:
> And pardon - forget about redefinition errors,

If you don't face the redefinitions errors anymore, situation is simple.
We can just add an ability to overwrite include path of stb/json by yourself by defining macros as you suggests.

Personally, I prefer

```
TINYGLTF_NO_INCLUDE_JSON
TINYGLTF_NO_INCLUDE_STB_IMAGE
TINYGLTF_NO_INCLUDE_STB_IMAGE_WRITE
```

Instead of 

```
TINYGLTF_BYPASS_INCLUDE_JSON
TINYGLTF_BYPASS_INCLUDE_STB_IMAGE
TINYGLTF_BYPASS_INCLUDE_STB_IMAGE_WRITE
```

How do you think?

kroko:
I think it's perfect! Thank you! 🙇

syoyo:
Awesome! Merged!

PR #149, opened 2019-03-05
glb: first commit

timknip:
Assumes a single buffer.

Guess this is not ready for merge, but its a start. Works for my use case.
```
// need set buffer byteLength
model.buffers[0].byteLength = model.buffers[0].data.size();

gltf_ctx.WriteGltfSceneToFile(&model, "tim.glb", false, true, true, true);
```

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -5188,6 +5188,7 @@ bool TinyGLTF::WriteGltfSceneToFile(Model *model, const std::string &filename,
   }
 
   if (writeBinary) {
+    model->buffers[0].byteLength = model->buffers[0].data.size();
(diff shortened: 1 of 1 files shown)


timknip:
Hmm. only tested this by loading a GLB.

Guess there will be issues when loading a non binary glTF and save it as GLB.

For example:
No clue how external / data uri textures are processed... Are these ending up in `buffer.data` ?

timknip:
Closing this PR, its too early.

syoyo:
> No clue how external / data uri textures are processed...

You'll need to set `embedImages` true

timknip:
@syoyo I'll do some testing with the Khronos test models and reopen the PR when all ok.

Issue #128, opened 2018-12-28
[TODO] Support Sparse Accessors


syoyo:
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#sparse-accessors

Ybalrid:
I was looking into this. It seems it's the recomended way of expressing morp animation targets within glTF. I'd like to load these from tiny_gltf. 

Initial support would simply require to actually expose the "sparse" metadata there:

 - https://github.com/syoyo/tinygltf/blob/master/tiny_gltf.h#L578

As I understand from khronos' spec:

- Sparse accessor works like a "patch" on the original data
- The "sparse" attribute contains 3 things :
    1) Number of elements changed in the original array
    2) BufferView (+ byte offset and type) of the "indices" in the array where data needs to be replaced
    3) BufferView (+ byte offset) of the "values" that needs to be read and put "in place" of each index pointed by the "indices" buffer view.

I'm interested into adding this to the accessor class in tinygltf : 

```cpp
struct {
  int count;
  struct {
    bool isSparse; 
    int byteOffset;
    int bufferView;
    int component_type; // a TINYGLTF_COMPONENT_TYPE_ value
 } indices;
  struct  {
    int bufferView;
    int byteOffset;
  } values;
} sparse;
```
And to add the code that fills these new data structure in the code that read the JSON, in case the "sparse" object is defined.

Using theses named anonymous structure, users will be able to get everything they need to read the updated array:

```cpp
const auto& the_accessor = model.accessors[some_index];
if(the_accessor.sparse.isSparse)
{
    //can access these members:
    the_accessor.sparse.count; //how many I need
    the_accessor.sparse.indices.bufferView; //where I find the indices
    the_accessor.sparce.indices.componentVype; //type of the indices
    the_accessor.sparce.indices.byteOffset; //where to start looking for indices
    the_accessor.sparce.values.bufferView; //where to find the new values to patch-in
    the_accessor.sparce.values.byteOffset; //where to start reading the new values
}
```
If I'm not mistaken, that's everything we need to do to make it possible to load a sparse accessor's data.

syoyo:
Good!

We need glTF files with sparse accessors for testing.

It looks there is only single glTF sample model in official glTF samples repo : https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/SimpleSparseAccessor

Ybalrid:
Yes, this file looks nice because it makes it obvious if it is working or not.

[...]

PR #150, opened 2019-03-06
Sparse accessor

Ybalrid:
- [x] Parsing of sparse accessor
- [x] loader-example updated to show the sparse accessor data 
- [x] glview example updated to show mesh moved by sparse data 

`loader-example` new output for : "SimpleSparseAccessor.gltf" :

```
  name         : 
    bufferView   : 0
    byteOffset   : 0
    componentType: UNSIGNED_SHORT(5123)
    count        : 36
    type         : SCALAR
    min          : [0]
    max          : [13]
  name         : 
    bufferView   : 1
    byteOffset   : 0
    componentType: FLOAT(5126)
    count        : 14
    type         : VEC3
    min          : [0, 0, 0]
    max          : [6, 4, 0]
    sparse:
      count  : 3
      indices: 
        bufferView   : 2
        byteOffset   : 0
        componentType: UNSIGNED_SHORT(5123)
      values : 
        bufferView   : 3
        byteOffset   : 0

```

See the "sparse:" in the 2nd accessor described correspound to what's expected.  https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/SimpleSparseAccessor/glTF/SimpleSparseAccessor.gltf#L75

~~I still need to cleanup and reformat tiny_gltf.hpp, and I'm looking into updtaing `glview`. It will need a little bit of refactoring because the way it loads buffers by parsing the buffer views is only usefull if OpenGL driver's has the `ARB_sparse_buffer` extension as far as I can tell. Simpler solution would be to only load buffers when parsing the accessors, and patch-in the sparse data that way.~~
> Edit, I've modified glview, and I ran clang-format on tiny_gltf. Please see comments below.

### DIFF
--- a/examples/glview/glview.cc
+++ b/examples/glview/glview.cc
@@ -28,7 +28,6 @@
 #include "tiny_gltf.h"
 #endif
 
-
 #define BUFFER_OFFSET(i) ((char *)NULL + (i))
 
 #define CheckGLErrors(desc)                                                   \
@@ -55,7 +54,9 @@ float eye[3], lookat[3], up[3];
 
 GLFWwindow *window;
 
-typedef struct { GLuint vb; } GLBufferState;
+typedef struct {
+  GLuint vb;
+} GLBufferState;
 
 typedef struct {
   std::vector<GLuint> diffuseTex;  // for each primitive in mesh
@@ -254,6 +255,26 @@ void motionFunc(GLFWwindow *window, double mouse_x, double mouse_y) {
   prevMouseY = mouse_y;
 }
 
+static size_t ComponentTypeByteSize(int type) {
+  switch (type) {
+    case TINYGLTF_COMPONENT_TYPE_UNSIGNED_BYTE:
+    case TINYGLTF_COMPONENT_TYPE_BYTE:
+      return sizeof(char);
+    case TINYGLTF_COMPONENT_TYPE_UNSIGNED_SHORT:
+    case TINYGLTF_COMPONENT_TYPE_SHORT:
+      return sizeof(short);
+    case TINYGLTF_COMPONENT_TYPE_UNSIGNED_INT:
+    case TINYGLTF_COMPONENT_TYPE_INT:
+      return sizeof(int);
+    case TINYGLTF_COMPONENT_TYPE_FLOAT:
+      return sizeof(float);
+    case TINYGLTF_COMPONENT_TYPE_DOUBLE:
+      return sizeof(double);
+    default:
+      return 0;
+  }
+}
+
 static void SetupMeshState(tinygltf::Model &model, GLuint progId) {
   // Buffer
   {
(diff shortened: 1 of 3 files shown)


Ybalrid:
Okay so, I've modified the glview example in the following way:

While loading a buffer, if the buffer is pointed to by a sparse accessor, it will copy the buffer into a temporary array, apply the replacement inside the array as described by the sparse accessor, then load this as the buffer.

If not, it will loaded without any modification.

2019-03-06-144305_1920x1080_scrot

Here's the test model to show that the sparse accessors are working.

Ybalrid:
This will fix #128

examples/glview/glview.cc
```diff
@@ -264,70 +285,173 @@ static void SetupMeshState(tinygltf::Model &model, GLuint progId) {
...
       std::cout << "buffer.size= " << buffer.data.size()
                 << ", byteOffset = " << bufferView.byteOffset << std::endl;
-      glBufferData(bufferView.target, bufferView.byteLength,
-                   &buffer.data.at(0) + bufferView.byteOffset, GL_STATIC_DRAW);
+
+      if (sparse_accessor < 0)
+        glBufferData(bufferView.target, bufferView.byteLength,
+                     &buffer.data.at(0) + bufferView.byteOffset,
+                     GL_STATIC_DRAW);
+      else {
```
Ybalrid: At this point, you could call this a hack because when glview was written, this wasn't supposed to be tied to an accessor, but it works.

loader_example.cc
```diff
@@ -409,6 +409,30 @@ static void Dump(const tinygltf::Model &model) {
         }
         std::cout << "]" << std::endl;
       }
+
+      if (accessor.sparse.isSparse) {
```
Ybalrid: Here is the only meaningful change to `loader_example`

tiny_gltf.h
```diff
@@ -575,7 +576,19 @@ struct Accessor {
   std::vector<double> minValues;  // optional
   std::vector<double> maxValues;  // optional
 
-  // TODO(syoyo): "sparse"
+  struct {
```
Ybalrid: here's where the sparse attributes are located inside the accessor

tiny_gltf.h
```diff
@@ -2804,6 +2833,51 @@ static bool ParseBufferView(BufferView *bufferView, std::string *err,
   bufferView->byteOffset = static_cast<size_t>(byteOffset);
   bufferView->byteLength = static_cast<size_t>(byteLength);
   bufferView->byteStride = static_cast<size_t>(byteStride);
+  return true;
+}
+
+static bool ParseSparseAccessor(Accessor *accessor, std::string *err,
```
Ybalrid: Here's the function that actually parse the "sparse" json object. It will be called by ParseAccessor only if it is needed.

syoyo:
Super awesome!!! 🎉🎉🎉🎉 Merged!

PR #154, opened 2019-03-07
Fix the display of the version history

Ybalrid:
Markdown is like HTML: simple line-breaks don't work.

Besides, making a list with dots here looks nicer :)


syoyo:
Awesome! Merged.

PR #155, opened 2019-03-21
Serialize accessor 'normalized' attribute.

ffreling:
When trying to generate a mesh from scratch and export a glTF file, I noticed that the `normalized` attribute of accessors is actually not exported.


syoyo:
Super awesome!!! Thanks!

PR #158, opened 2019-04-25
 Add checks for boundary conditions for malformed glTF files

jwmcglynn:
When loading untrusted glTF files, ideally an error should be returned if the file is malformed instead of an exception/crash.  Add additional validation for crashes found when running tinygltf under a fuzzer, and add test cases to confirm functionality.

Additionally, set up tests to run on travis-ci.


syoyo:
Super awesome! Thank you!

PR #160, opened 2019-04-27
Parse integers directly instead of casting doubles

jwmcglynn:
When parsing numeric values as doubles, its possible for users to
specify values that cannot be converted to integers, such as Inf, NaN,
and extremes such as 1e100.  If this value is received, and then cast to
an int, it is undefined behavior, which trips ubsan when running
tinygltf under a fuzzer.

Instead of parsing integral values as doubles, use nlohmann/json's
built-in support to parse integer and unsigned values directly, with
.is_number_integer() and .is_number_unsigned().

Add ParseIntegerProperty, ParseUnsignedProperty, and
ParseIntegerArrayProperty helpers that allow parsing directly to
int/uint values and update code to use them when appropriate.

### DIFF
--- a/models/BoundsChecking/integer-out-of-bounds.gltf
+++ b/models/BoundsChecking/integer-out-of-bounds.gltf
@@ -1,0 +1,67 @@
+{
+  "scenes": [
+    {
+      "nodes": [0]
+    }
+  ],
+  "nodes": [
+    {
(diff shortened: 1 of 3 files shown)


syoyo:
Super awesome! This makes tinygltf parser more robust. Thanks!

PR #161, opened 2019-04-29
if image.uri empty, should use image.bufferView

rainliang000:
if image.uri empty, shoud use image.mimeType and image.bufferview


syoyo:
Awesome! Merged.

PR #163, opened 2019-05-03
Add "asset" section to BoundsChecking test models

jwmcglynn:
These test assets were trimmed down and erroneously didn't have an
"asset" section or version number, which is required by the glTF spec.

This fixes test failures with pull request #162.


syoyo:
Awesome! merged.

PR #169, opened 2019-05-25
Remove Mesh::targets, as targets only exists inside `Primitive`

Ybalrid:
In a glTF asset, only mesh primitives should have a a "targets" property.

This PR remove the unused data structure and parsing code for this non-existent storage for target metadata.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -708,7 +708,6 @@ struct Mesh {
   std::string name;
   std::vector<Primitive> primitives;
   std::vector<double> weights;  // weights to be applied to the Morph Targets
-  std::vector<std::map<std::string, int> > targets;
   ExtensionMap extensions;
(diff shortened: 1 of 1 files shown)


tiny_gltf.h
```diff
@@ -4912,14 +4892,13 @@ static void SerializeGltfBufferView(BufferView &bufferView, json &o) {
 }
 
 static void SerializeGltfImage(Image &image, json &o) {
- 	// if uri empty, the mimeType and bufferview should be set
```
Ybalrid: This is just the result of running `clang-format` on the file. apparently this was forgotten after making these changes.

syoyo:
Awesome! merged.

I think when I implemented it(before the glTF 2.0 spec freeze), it was defined in `mesh`.

PR #170, opened 2019-06-15
Fix C4267 warning (size_t -> int) convertion

Ybalrid:
Everthing's in the title.

With this tiny change, tinygltf builds 100% warning free on MSVC 2019.


syoyo:
Thanks! Merged.

PR #171, opened 2019-06-23
Material names are stored once in material.name, then duplicated in "additionalValues"

Ybalrid:
This patch prevent this duplication.

Without this patch, you have both Material.name being filled in by the caller function, and then this function insert "name" in the map that only should *only* contains the alpha blending and normal/occlusion/emissive values.


Ybalrid:
Also, I checked, `material.name` is also serialized independently, so we are all good on that front already!

syoyo:
Awesome! Merged!

PR #173, opened 2019-07-04
Adding texture scale reading + asset copyright and texture names to the export

christophe-f8:
Added the reading of texture scales (those are used for normal maps).

Added the asset copyright and texture names to the output files while saving

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -393,6 +393,17 @@ struct Parameter {
     return 0;
   }
 
+  /// Return the scale of a texture if this Parameter is a normal texture map.
+  /// Returned value is only valid if the parameter represent a normal texture
+  /// from a material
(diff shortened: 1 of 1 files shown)


syoyo:
Super awesome! Thank you!

christophe-f8:
You're very welcome. Thanks for this library :-)

Ybalrid:
Hey, I wasn't even aware this property existed. I just checked the spec. We are also missing the `strength`  parameters of occlusion maps. It needs to be added to `tinygltf`. It pretty much serve the same purpose as the normal texture scale, except it's a scalar between [0.0; 1.0]

@syoyo I'll send a PR :wink:

christophe-f8:
Thank you, I'll probably use this property as well in the future

PR #174, opened 2019-07-04
 Add Parameter::TextureStrength(). Precise default values. 

Ybalrid:
The default values on these methods as been set to what is described [here](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0?ts=4#normaltextureinfo) and [here](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0?ts=4#occlusiontextureinfo).

This is to keep consistent with the API behavior from #144

(also ran clang-format for good measure)

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -390,6 +390,7 @@ struct Parameter {
     if (it != std::end(json_double_value)) {
       return int(it->second);
     }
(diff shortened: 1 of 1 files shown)


syoyo:
Good catch. Merged!

PR #185, opened 2019-08-13
remove extra #endif in examples/basic app

DerouineauNicolas:
Hi,
I get the following error compiling on macOS:

    examples/basic/main.cpp:336:2: error: #endif without #if

This PR removes the useless #endif.


syoyo:
Thanks!

PR #191, opened 2019-08-16
Animation extension properties

Selmar:
adding extensions property to `Animation` and `AnimationChannel`


syoyo:
Awesome! Merged!

PR #193, opened 2019-08-19
Emissive factor defaults

Selmar:
Default material array values were only set for `baseColorFactor`, not for `emissiveFactor`. This caused a mismatch between material data after writing and parsing again.

This PR sets the default values for `emissiveFactor` as well.

Question: I noticed that there is currently some inconsistency between which default values are always set to defaults and/or serialized and which aren't. For example, node translation/rotation/scale is just parsed/serialized as-is, ignoring defaults. What is your preferred approach?

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -3743,8 +3743,22 @@ static bool ParsePbrMetallicRoughness(PbrMetallicRoughness *pbr,
 static bool ParseMaterial(Material *material, std::string *err, const json &o) {
   ParseStringProperty(&material->name, err, o, "name", /* required */ false);
 
-  ParseNumberArrayProperty(&material->emissiveFactor, err, o, "emissiveFactor",
-                           /* required */ false);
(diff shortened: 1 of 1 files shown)


syoyo:
Thanks!

> I noticed that there is currently some inconsistency between which default values are always set to defaults and/or serialized and which aren't. For example, node translation/rotation/scale is just parsed/serialized as-is, ignoring defaults. What is your preferred approach?

This may be a difficult question.

For a short-term perspective, I'd prefer ignore default values when serializing(this will save .gltf file size) and set default values when there is no corresponding entries in glTF at parsing.

For a mid-term perspective, it would be better to introduce an option whether ignore/not-set or export/set default values in serializer/parser, since default value may be changed in the future glTF spec(by minor updating) and some people may want force output full description of glTF data in the serializer.

For a long-term perspective, we are better to transit to automatically generate parser/serializer based on JSON schema to avoid human errors > https://github.com/syoyo/tinygltf/issues/10

PR #199, opened 2019-08-29
GetTypeSizeInBytes change to GetNumComponentsInType

DingboDingboDingbo:
Changed name to be less misleading, as GetTypeSizeInBytes was not returning Type Size In Bytes


syoyo:
Awesome! merged with small fix(rename a variable)

PR #201, opened 2019-09-03
RapidJSON, More Move Semantics, C++14 optional/off

jrkoonce:


jrkoonce:
These changes include:
1. Move support
  A. tinygltf objects now support move semantics and are moved as often as possible
  B. json objects are moved as often as possible instead of copied
2. RapidJSON 1.1.0 support (#define TINYGLTF_USE_RAPIDJSON)
3. Compilation on MSVC, gcc, and clang using C++11
4. Optional support for C++14 (#define TINYGLTF_USE_CPP14)
5. Unit tests (tester.cc) compile and pass using nlohmann or RapidJSON either one

syoyo:
Awesome!

Still, it has some warnings when compiling `examples/glview`

```
In file included from /home/syoyo/work/tinygltf/examples/glview/glview.cc:28:
/home/syoyo/work/tinygltf/examples/glview/../../tiny_gltf.h:1187:10: warning: explicitly defaulted copy assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
  Model &operator=(const Model &) = default;
         ^
/home/syoyo/work/tinygltf/examples/glview/../../tiny_gltf.h:1234:9: note: copy assignment operator of 'Model' is implicitly deleted because field 'asset' has a deleted copy assignment operator
  Asset asset;
        ^
/home/syoyo/work/tinygltf/examples/glview/../../tiny_gltf.h:1111:3: note: copy assignment operator is implicitly deleted because 'Asset' has a user-declared move constructor
  Asset(Asset&& rhs) noexcept
  ^
```

jrkoonce:
Ahh got it. I’ll fix it. 

Thanks!

Rusty

jrkoonce:
The default RapidJSON implementation uses a pool allocator which is great for speed and minimal memory fragmentation; however, it is not thread safe and requires only a single JSON load/save is active at once.  I added support for their CRT allocator which is stateless and thread safe.  Now if you define #TINYGLTF_USE_RAPIDJSON_CRTALLOCATOR it will be used and loading/saving multiple gltf's simultaneously is supported at the cost of more memory fragmentation and a lot more allocation/deallocation overhead.  There is an assertion when more than one JsonDocument is created at once telling you to define TINYGLTF_USE_RAPIDJSON_CRTALLOCATOR if you need that functionality.

syoyo:
Thanks! Merged(There was still few compiler warnings. I suppressed it).

> TINYGLTF_USE_RAPIDJSON_CRTALLOCATOR

I see.

PR #211, opened 2019-10-18
small serialization additions

Selmar:
- add some missing serialization
- add mesh==() weights comparison
- use const iterator for extension serialization


Selmar:
+ correctly deserialize materials without pbrMetallicRoughness

syoyo:
Awesome! merged.

Issue #197, opened 2019-08-26
Move Constructors


jrkoonce:
The library often utilizes vectors of user defined structs/classes.  None of those structs/classes are defining move constructors, leaving them to be implicitly generated by the compiler if possible.  From first glance, it appears most of them can be implicitly generated, however, std::vector also has the requirement that the move constructor be declared noexcept before it can be used.

I am not sure implicitly generated move constructors will be noexcept .  The move constructor can be explicitly defined to be the default implementation and noexcept both: "T(T&&) noexcept = default;" This ensures the move constructor can be implicitly generated by the compiler and that it is noexcept, or it won't compile.  As coded now, it will compile and be a silent copy instead of move whenever the non-throwing move constructor can't be generated.

I would add the definition of the defaulted explicit noexcept move constructors, to ensure maximum CPU and memory efficiency, with compile errors indicating when a copy would occur instead of a move.  When a compile error occurs, an explicit noexcept move constructor can be programmed.  Thus ensuring all user defined structs/classes will be correctly moved and not copied inside of std::vector.

syoyo:
tinygltf was initially written for targeting C++03 compiler, and there is no move constructors used in current `master` yet. 

PR of adding wide support of move semantics including move constructor to tinygltf classes is appreciated.

jrkoonce:
I made a fork and have added them. Reduces allocations by a lot. I’ll check it in today and you can integrate it back if you want. It is a pretty quick implementation, so you will want to check it across all compilers and clean it up.

Rusty

syoyo:
Code looks nice. Please create a PR. It will trigger CI build on Linux.

jrkoonce:
Well, I have good news and bad news.  Good news, is I got a lot more move semantics in and I added support for RapidJSON 1.1.  Bad news is I used a C++14 feature (auto parameter to lambda) not realizing you restricted the code to C++11.  I can rewrite the code to not use that feature, but it makes the code much more verbose and long.   The RapidJSON support was rushed and probably could use a bit more cleanup and definitely more testing.

syoyo:
Awesome! > RapidJSON support.

>  I used a C++14 feature (auto parameter to lambda) not realizing you restricted the code to C++11

We still need to stick with C++11. Could you rewrite the code in C++11?
(We may switch to use C++14 in the future release(and it will bump the major version(i.e. v3.x))

jrkoonce:
[...]

Issue #50, opened 2018-03-15
Buffer.uri not set after loading buffer


Selmar:
This seems inconsistent with loading images. In addition, we actually need the uri for caching purposes.

Is there a specific reason for this, or can we perhaps save it? : )

syoyo:
We may store add original url to `Buffer.orignal_url`  if required.
(Not sure its ok to save url to `Buffer.url`)

> we actually need the uri for caching purposes.

How about to use `Buffer.name` for caching?

Selmar:
We require the file to get the timestamp.

Currently the `buffer.uri` property is not referenced anywhere.

Adding `original_uri` is fine, too, of course.

syoyo:
Ah, it should be simply my mistake not saving uri.

Now fixed here: https://github.com/syoyo/tinygltf/commit/20244e1e4e8938b2e51d109a4bc2f57946e5ee5d#diff-f70622f14b85ce3409f1f100e5c47ec4R1789

PR #212, opened 2019-10-21
Move semantics

Selmar:
Implement default noexcept move constructor/operator, as discussed in #197 

I don't have a C++11 compiler available, but godbolt seemed happy.

@jrkoonce I guess this works, then?

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -271,23 +271,10 @@ class Value {
   explicit Value(const Object &o) : type_(OBJECT_TYPE) { object_value_ = o; }
   explicit Value(Object &&o) noexcept : type_(OBJECT_TYPE),
                                         object_value_(std::move(o)) {}
-  Value(Value &&rhs) noexcept : type_(rhs.type_),
-                                int_value_(rhs.int_value_),
-                                real_value_(rhs.real_value_),
-                                string_value_(std::move(rhs.string_value_)),
-                                binary_value_(std::move(rhs.binary_value_)),
-                                array_value_(std::move(rhs.array_value_)),
-                                object_value_(std::move(rhs.object_value_)),
-                                boolean_value_(rhs.boolean_value_) {}
-  Value(const Value &rhs) = default;
-  Value &operator=(const Value &rhs) = default;
-  Value &operator=(Value &&rhs) {
-    if (this != &rhs) {
-      this->~Value();
-      new (reinterpret_cast<void *>(this)) Value(std::move(rhs));
-    }
-    return *this;
-  }
+  Value(const Value &) = default;
+  Value(Value &&) noexcept = default;
+  Value &operator=(const Value &) = default;
+  Value &operator=(Value &&) noexcept = default;
 
   char Type() const { return static_cast<const char>(type_); }
 
@@ -498,6 +485,11 @@ struct Parameter {
          (number_array.size() > 3 ? number_array[3] : 1.0)}};
   }
 
+  Parameter() = default;
+  Parameter(const Parameter &) = default;
+  Parameter(Parameter &&) noexcept = default;
+  Parameter &operator=(const Parameter &) = default;
+  Parameter &operator=(Parameter &&) noexcept = default;
   bool operator==(const Parameter &) const;
 };
 
@@ -522,6 +514,10 @@ struct AnimationChannel {
   ExtensionMap extensions;
 
   AnimationChannel() : sampler(-1), target_node(-1) {}
+  AnimationChannel(const AnimationChannel &) = default;
+  AnimationChannel(AnimationChannel &&) noexcept = default;
+  AnimationChannel &operator=(const AnimationChannel &) = default;
+  AnimationChannel &operator=(AnimationChannel &&) noexcept = default;
   bool operator==(const AnimationChannel &) const;
 };
 
@@ -533,6 +529,10 @@ struct AnimationSampler {
   Value extras;
 
   AnimationSampler() : input(-1), output(-1), interpolation("LINEAR") {}
+  AnimationSampler(const AnimationSampler &) = default;
+  AnimationSampler(AnimationSampler &&) noexcept = default;
+  AnimationSampler &operator=(const AnimationSampler &) = default;
+  AnimationSampler &operator=(AnimationSampler &&) noexcept = default;
   bool operator==(const AnimationSampler &) const;
 };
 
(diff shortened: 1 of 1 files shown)


Selmar:
So apparently compilation success depends on the gcc version. I guess this is the issue you ran into, @jrkoonce .

Selmar:
probably relevant:

https://stackoverflow.com/questions/20610658/defaulted-move-assignment-cannot-be-explicitly-noexcept-if-a-member-has-a-non-tr

syoyo:
Travis error happens when compiling unit test with clang++.

When I modify travis script to show clang version,

```
0.28s$ clang++ -v
clang version 7.0.0 (tags/RELEASE_700/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/clang-7.0.0/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/4.8
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/4.8.5
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/4.9
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/4.9.4
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/5
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/5.4.0
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/6
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/6.0.0
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/5.4.0
Candidate multilib: .;@m64
Selected multilib: .;@m64
```

it looks compilation error is due to using gcc 4.8 headers(gnustl).
When I compile unit tester with gcc 4.9 STL headers or `-std=libc++`, error does not happen.

So, the code itself should be ok.

syoyo:
Compilation error happens with gcc 4.8. I can reproduce it with CentOS7's default gcc(4.8)

```
$ g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --enable-languages=c,c++,objc,obj-c++,java,fortran,ada,go,lto --enable-plugin --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.8.5 20150623 (Red Hat 4.8.5-39) (GCC) 
$ make
g++   -std=c++11 -g -O0 -o loader_example loader_example.cc 
In file included from loader_example.cc:7:0:
tiny_gltf.h:275:3: error: function ‘tinygltf::Value::Value(tinygltf::Value&&)’ defaulted on its first declaration with an exception-specification that differs from the implicit declaration ‘tinygltf::Value::Value(tinygltf::Value&&)’
   Value(Value &&) noexcept = default;
   ^
tiny_gltf.h:277:10: error: function ‘tinygltf::Value& tinygltf::Value::operator=(tinygltf::Value&&)’ defaulted on its first declaration with an exception-specification that differs from the implicit declaration ‘tinygltf::Value& tinygltf::Value::operator=(tinygltf::Value&&)’
   Value &operator=(Value &&) noexcept = default;
          ^
```

It is known that gcc 4.8 does not fully support C++11. 
Still, I'd like to support old compiler since sometime we have to stick to use gcc 4.8(e.g. compile glTF Maya plugin on CentOS7 platform). Possible solution is to use ifdef's.

jrkoonce:
Visual Studio 2017 compiles the changes fine.  Last I read, it depends on if the STL implementation defines a particular method we are using on std::vector as noexcept or not, and vendor's are allowed to define it either way.  Older versions that don't have it noexcept cause the problem I think.  OR I'm completely fool of s*it. :)

Simplest solution may be to use a macro NOEXCEPT and have it defined to "noexcept" when you enable the exception specification, and defined to nothing when you don't, then set it based on STL implementation or compiler version.  And of course replace all "noexcept" with "NOEXCEPT" macro.

I am usually not a macro fan, but you can also make a macro to explicitly define the copy/move/assignment operators as default impls.  I would personally add the destructor to it so they are also explicitly defined as defaulted too.  I'd do that so if someone adds a destructor it wouldn't compile, or they at least get a hint that defining the destructor would have unintended consequences.

#if USE_NOEXECEPT
#define NOEXCEPT noexcept
#else
#define NOEXCEPT
#50 
#define DEFAULT_METHODS(x) ~x() = default; x(const x&) = default; x(x&&) NOEXCEPT = default; ...

class Value
{
public:
DEFAULT_METHODS(Value)
};

syoyo:
> Older versions that don't have it noexcept cause the problem I think.

For gcc, according to gcc4.8's C++11 status

https://gcc.gnu.org/gcc-4.8/cxx0x_status.html

> Allowing move constructors to throw [noexcept] | N3050 | Yes (core language only)

`noexcept` support is only in language level, so GNU STL implementation is not complete(it may have different signature without `noexcept`)

syoyo:
@Selmar thanks!

> Travis says the previously failing build it is using the gcc compiler version 4.9

Actually, the failure was caused in clang7 + gcc 5.4 STL combination.

https://travis-ci.org/syoyo/tinygltf/jobs/601353390#L479

So there was another reason for the compilation error.
I have also confirmed the build gets ok with gcc 4.8(CentOS7)

After some extra tests(e.g. build `basic` viewer), I will merge this PR.

Selmar:
> Actually, the failure was caused in clang7 + gcc 5.4 STL combination.
> 
> https://travis-ci.org/syoyo/tinygltf/jobs/601353390#L479
> 
> So there was another reason for the compilation error.

I don't understand why this one wouldn't still fail, then.

syoyo:
Tested with some examples and works file, so merged.

PR #216, opened 2019-10-30
Feature/utf8 path

harokyang:
Now all filepath that tinygltf API use will be utf8 encoded std::string

On Windows, those string will be convertd to wstring before used on file operation
On Linux, it is utf8 by default and nothing changed

Based on gltf 2.0 spec, all path of resources described in gltf will also be utf8 string
This merge also make any internal resource finding support utf8 path

On user side, make sure to convert any local filepath to utf8 encoded string(highly recommanded with filesystem::path::u8string() )


syoyo:
Awesome! Merged!

PR #226, opened 2019-11-18
Glb bin blob

epajarre:
I am very new to Github workflow, so I hope I got this correct, and  you check it anyways.


syoyo:
Thanks! merged!

PR #230, opened 2019-12-18
Fix the issue that the extension of primitive written as extension of mesh.

itainter:


syoyo:
Super awesome! thank you!

Issue #116, opened 2018-12-05
Bufferview target error, zero when it should be element array enum


viperscape:
Given the spec: https://raw.githubusercontent.com/KhronosGroup/glTF/master/specification/2.0/schema/mesh.primitive.schema.json
> Primitive Indices: "The index of the accessor that contains mesh indices.  When this is not defined, the primitives should be rendered without indices using `drawArrays()`.  When defined, the accessor must contain indices: the `bufferView` referenced by the accessor should have a `target` equal to 34963 (ELEMENT_ARRAY_BUFFER); `componentType` must be 5121 (UNSIGNED_BYTE), 5123 (UNSIGNED_SHORT) or 5125 (UNSIGNED_INT), the latter may require enabling additional hardware support; `type` must be `\"SCALAR\"`. For triangle primitives, the front face has a counter-clockwise (CCW) winding order."

The loader states that some gltf files have mesh primitives with a bufferview.target of 0, however see attached rock.gltf file, it states indices of 4, which means the target should be 34963. From the spec, unless I'm misunderstanding it, states that the target should only ever be 34963 or 34962 (when indices is missing). Another example is the Khronos avocado gltf2.0 example mesh

[rock.zip](https://github.com/syoyo/tinygltf/files/2649745/rock.zip)

viperscape:
Just to add to this, forcing the target to be Element Array Buffer renders the mesh properly.
```c++
int target = bufferView.target == 0 ? GL_ELEMENT_ARRAY_BUFFER : bufferView.target;
glBindBuffer(target, vbo);
glBufferData(target, bufferView.byteLength,
    &buffer.data.at(0) + bufferView.byteOffset, GL_STATIC_DRAW);
```
So the problem lies specifically with the parser for 'target', I think, as the buffers are fine

syoyo:
Thanks!

Even though it looks the spec does not state **default** `target` value, it looks its safe to initialize it as 34963 (ELEMENT_ARRAY_BUFFER)

> Just to add to this, forcing the target to be Element Array Buffer renders the mesh properly.

You can send a PR!

syoyo:
I thinks this issue is ready to close thanks to the merge of https://github.com/syoyo/tinygltf/pull/117

viperscape:
Yes thanks, just saw this. Looks like after the recent merge, everything is fine

PR #232, opened 2020-01-02
Do not set target on bufferView pointing to animation data

kacprzak:
I was implementing animations and creating OpenGL buffers based on BufferView `target` value. It is nice that tinygltf sets target correctly for Indices and Attributes, but BufferView can point to other type of data that not necessarily will land in GPU buffer. Probably that is why `target` value in gltf specification is optional.

https://github.com/syoyo/tinygltf/issues/116#issuecomment-444569685
After reading this I can add that this was maybe safe but not very useful. I propose to leave target value as 0 for non vertex indices and non vertex attribute data.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -5526,7 +5526,7 @@ bool TinyGLTF::LoadFromString(Model *model, std::string *err, std::string *warn,
 
   // Assign missing bufferView target types
   // - Look for missing Mesh indices
-  // - Look for missing bufferView targets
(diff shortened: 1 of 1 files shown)


syoyo:
Thanks! Merged!

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#bufferviewtarget

>  I propose to leave target value as 0 for non vertex indices and non vertex attribute data.

I see. it will make sense.

PR #235, opened 2020-01-07
BufferView ctor now uses correct members order.

ziocleto:
tiny_gltf.h line 815 
(Moved target(0) to second to last)


syoyo:
Thanks!

Issue #236, opened 2020-01-13
Images with spaces do not load on Linux


FrankGalligan:
I attached a sample glTF file that has an external reference to an image with spaces in the filename.

Below is the text for the images element in the glTF file:
"images": [
    {
      "uri": "2x2 image has spaces.png"
    }
  ],

[CubeImageUriSpaces.zip](https://github.com/syoyo/tinygltf/files/4056395/CubeImageUriSpaces.zip)

FrankGalligan:
Per this https://github.com/KhronosGroup/glTF/issues/1449 spaces are valid.

syoyo:
Thanks! PR has been merged.

FYI,  percent-encoded URI support will be soon merged: https://github.com/syoyo/tinygltf/issues/228

AlexApps99:
@syoyo @FrankGalligan 
Now this repo can't be cloned on Windows at all :(
```
error: invalid path 'models/CubeImageUriSpaces/ 2x2 image  has multiple      spaces.png '
Unable to checkout 'bcf2ce586ee8bf2a2a816afa6bfe2f8692ba6ac2' in submodule path 'lib/tinygltf'
```

syoyo:
@AlexApps99 I can git clone using Git for Windows(MinTTY) on Windows 10. what git configuration do you use?

AlexApps99:
I included it as a submodule in my engine and my CI (github actions) spat out that error cloning with https://github.com/actions/checkout

AlexApps99:
It could be a filesystem error, as CI may not use NTFS like traditional windows installations

AlexApps99:
Nevermind, the issue seems to be that the file extension ends with spaces (`.png `)

syoyo:
At least filename with whitespace should be handled well on Windows(at least AppVeyor CI build works well https://ci.appveyor.com/project/syoyo/tinygltf )

`git config core.protectNTFS false` may help. 

https://stackoverflow.com/questions/29294910/unable-to-add-files-with-name-containing-tilde-followed-by-a-number/29300360#29300360

AlexApps99:
Since the file does not have a correct png file extension, it shouldn't even be a part of the tests.
I can't change the system configuration as it is CI and resets every commit.

AlexApps99:
The problem isn't that there is whitespace, but that it is at the start and end of the filename.
I don't know how or why, but it doesn't work at all with GitHub Actions (which me and a lot of other projects use)

[...]

PR #237, opened 2020-01-13
Fix loading images with spaces on Linux

FrankGalligan:
This change quotes the string before it is passed into wordexp.

This addresses issue https://github.com/syoyo/tinygltf/issues/236


syoyo:
Awesome! Thanks!

AlexApps99:
See comment here:
https://github.com/syoyo/tinygltf/issues/236#issuecomment-575497280

PR #238, opened 2020-01-17
Initialize defaultScene to -1

sammyKhan:
To prevent undefined behavior if the model is serialized without defaultScene being set explicitly.


syoyo:
Super awesome! Thanks!

PR #239, opened 2020-01-21
Channel target extensions

Selmar:
The [`animation.channel.target`](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#channeltarget-white_check_mark) object is a fully-qualified gltf object and as such should support extensions.

Added as `ExtensionMap target_extensions;` inside the `AnimationChannel` struct, because I did not want to break existing implementations. Is it acceptable for you?


syoyo:
Awesome! Merged!

> because I did not want to break existing implementations

Better solution will be introducing `AnimationChannelTarget` class,  but this would break existing API. So we are better to consider it in the next major release.

https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/animation.channel.target.schema.json

PR #240, opened 2020-01-22
Set target for bufferView used by MorphTargets

kacprzak:
When setting target to ARRAY_BUFFER only for attributes we forgot about attributes of morph targets.


syoyo:
Awesome! Merged!

PR #242, opened 2020-01-28
Some empty array fixes

Selmar:
dont serialize some arrays if they're empty: 
- bufferviews
- accessors
- buffers

the diff makes it seem like i changed a lot but it's mostly whitespace changes.. ^^'


syoyo:
Super awesome! :tada:

Merged!

PR #243, opened 2020-02-20
Reading copyright field in ParseAsset()

mfpgs:
Added missing copyright read in the ParseAsset() function.


syoyo:
Awesome! Thanks!

PR #247, opened 2020-02-20
Avoid duplicating path string in ExpandFilePath

moretromain:
Changine signature of `ExpandFilePath()` to avoid extra string copies/allocations

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -1203,7 +1203,7 @@ typedef bool (*FileExistsFunction)(const std::string &abs_filename, void *);
 ///
 /// ExpandFilePathFunction type. Signature for custom filesystem callbacks.
 ///
-typedef std::string (*ExpandFilePathFunction)(const std::string &, void *);
+typedef void (*ExpandFilePathFunction)(std::string *, void *);
 
 ///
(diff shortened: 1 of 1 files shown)


syoyo:
Changing API will affect programs using tinygltf, so I don't want to merge this change unless this change give significant performance improvement(but looks no)

moretromain:
I understand this affects the API and might break some existing code. On our end, we use a custom set of FSCallbacks, and our `ExpandFilePath()` doesn't do anything. But with the actual signature, we still need to return an extra copy of the input argument.
We're in a context where we really need to minimize allocations and string copies as much as we can, so I thought this would be a welcome change. I understand if you don't want to break the current API though..
Thanks!

tiny_gltf.h
```diff
@@ -1203,7 +1203,7 @@ typedef bool (*FileExistsFunction)(const std::string &abs_filename, void *);
 ///
 /// ExpandFilePathFunction type. Signature for custom filesystem callbacks.
 ///
-typedef std::string (*ExpandFilePathFunction)(const std::string &, void *);
```
Snowapril, line 1206: .

PR #249, opened 2020-03-05
Ensures only valid bufferViews are mentioned in accessor

sanjeetsuhag:
This change allows specification of an accessor without a bufferView and checks that bufferView index in an accessor is not negative.


syoyo:
Super awesome! Thanks!

PR #250, opened 2020-03-09
Prevent cloning of images in buffers

FsiGuy00015623:
Now, in UpdateImageObject, if an image has no URI but had a valid bufferView, it will no longer change the image object or copy the buffer object to an outside file at baseDir. 
(When saving files from streams, this function would previously save textures in buffers as "#.png" in the program's working directory instead of the model file's directory, and create a broken link in the updated image object if the two locations were different.)


syoyo:
Awesome! Thanks!

Merged!

PR #252, opened 2020-04-09
Fixed small error if node has no mesh in examples/basic.

freddiehonohan:
See title and thanks for the library!


syoyo:
Super awesome! Thanks!

PR #254, opened 2020-04-12
Fix write functions so they work with MinGW

### LINKED ISSUE #253: SerializeGltfBufferData fails when compiling with msys2+mingw on Windows
In `SerializeGltfBufferData`, `_wopen` always returns -1 for me.

If I replace
```C++
#if defined(__GLIBCXX__)  // mingw
  int file_descriptor =
      _wopen(UTF8ToWchar(binFilename).c_str(), _O_WRONLY | _O_BINARY);
  __gnu_cxx::stdio_filebuf<char> wfile_buf(file_descriptor, std::ios_base::in);
  std::ostream output(&wfile_buf);
  if (!wfile_buf.is_open()) return false;
#elif defined(_MSC_VER)
  std::ofstream output(UTF8ToWchar(binFilename).c_str(), std::ofstream::binary);
  if (!output.is_open()) return false;
#else
```
with just
```C++
#if defined(__GLIBCXX__) || defined(_MSC_VER)
  std::ofstream output(UTF8ToWchar(binFilename).c_str(), std::ofstream::binary);
  if (!output.is_open()) return false;
#else
```
it works just fine (even with UTF-8 filenames). It requires C++17, however.

LHLaurini:
Fixes #253

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -2660,8 +2660,11 @@ bool WriteWholeFile(std::string *err, const std::string &filepath,
 #ifdef _WIN32
 #if defined(__GLIBCXX__)  // mingw
   int file_descriptor =
-      _wopen(UTF8ToWchar(filepath).c_str(), _O_WRONLY | _O_BINARY);
-  __gnu_cxx::stdio_filebuf<char> wfile_buf(file_descriptor, std::ios_base::in);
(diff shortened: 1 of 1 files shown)


syoyo:
Awesome! Merged!

PR #266, opened 2020-06-04
fix an error occurred while expanding utf8 path

cwbhhjl:
ExpandEnvironmentStringsA can't handle utf8 string correctly, even I have change code page to utf8.

test code
```
setlocale(LC_ALL, ".65001");
std::string s1 = u8"对";  // e5 af b9
char buffer[255];
ExpandEnvironmentStringsA(s1.c_str(), buffer, 255);

// buffer : e5 af 00 00 .........
```

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -2483,6 +2483,15 @@ static inline std::wstring UTF8ToWchar(const std::string &str) {
                       (int)wstr.size());
   return wstr;
 }
+
+static inline std::string WcharToUTF8(const std::wstring &wstr) {
+  int str_size = WideCharToMultiByte(CP_UTF8, 0, wstr.data(), (int)wstr.size(),
+                                      nullptr, 0, NULL, NULL);
+  std::string str(str_size, 0);
+  WideCharToMultiByte(CP_UTF8, 0, wstr.data(), (int)wstr.size(), &str[0],
+                      (int)str.size(), NULL, NULL);
+  return str;
+}
 #endif
 
 #ifndef TINYGLTF_NO_FS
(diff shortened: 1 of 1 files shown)


syoyo:
Which source code file encoding are you using(MBCS or UNICODE)?

Probably we may need to use `ExpandEnvironmentStringsA` or `ExpandEnvironmentStringsW` depending on source file encoding(using `UNICODE` ifdef)

cwbhhjl:
> 
> 
> Which source code file encoding are you using(MBCS or UNICODE)?
> 
> Probably we may need to use `ExpandEnvironmentStringsA` or `ExpandEnvironmentStringsW` depending on source file encoding(using `UNICODE` ifdef)

I use UNICODE and I think MBCS has the same problem.

WIN32 API has define the macro `ExpandEnvironmentStrings`  depend on `UNICODE`.  I think it would be better if we assume all strings(path/uri in gltf) are utf8 and then convert string to wstring and call the _W API.

syoyo:
I have tested the code with VS2019(Use utf-8 as an file encoding and UNICODE as a character set for MSVC compiler) and ExpandEnvironmentStringsA returns  ` e5 af b9`.

I have also tested the code with some modification

- Use S-JIS encoding(common multibyte character encoding in Japanese) since I don't know what is common encoding for Chinese multibyte characters
- Use Japanese character "対"
- Both use Unicode and MBCS when compiling

`ExpandEnvironmentStringsA` still works well.

utftest

@cwbhhjl Could you please write an unit test to reproduce this issue? Here is an example unit code(in tinyexr repo) : https://github.com/syoyo/tinyexr/blob/master/test/unit/tester.cc#L82

cwbhhjl:
@syoyo I did some tests and found strange results. 

https://rextester.com/RIJZM69582  online code, the result is correct.

Then I compiled an exe with the same code.

On my computer(Win10 Pro - simplified chinese) : wrong
批注 2020-06-05 134207

On hyperV Win10 Pro - English: correct
winen_result

Then I tested it on several different computers (all of them are win10 simplified chinese), all results were wrong, the last byte was missing.

syoyo:
@cwbhhjl Could you try setting the encoding of console to UTF-8(Using `cmd.exe` may be easier to try it on than PowerShell) in Simplified Chinese Win 10 environment and how it goes?

https://superuser.com/questions/269818/change-default-code-page-of-windows-console-to-utf-8

Probably system locale affects the locale and/or Win32 API behaviour of .exe binary.

cwbhhjl:
@syoyo I found `chcp 65001` does not affect the result. 
I probably tested the options that affect the result.  open Control Pannel -> time & region -> region -> administrative -> language for non unicode program -> change system local...
If I set to English or Japanese, the result of the program will become correct.
If I set to Chinese(traditional or simplified), the result of the program will become wrong.

Is it convenient for you to try this test?

cwbhhjl:
In addition, if I set to chinese and select the utf8 beta option, the result will also be correct.

syoyo:
@cwbhhjl Thanks! I can reproduce the issue with Chinese codepage(936). The binary is compiled with MBCS charset encoding.

chaencoding

So, in any case, we'll better to use wchar and `W` suffix API  for string where UTF-8 input is expected.  

Basically, your patch would work both for MBCS and UNICODE binary on Chinese and UTF-8 locale encodings, but let me give some time to verify it.

syoyo:
I have added a unit test and verified the PR works well on  Windows 10 + codepage 936(tested both MSVC MBCS and UNICODE build) etc...

https://github.com/syoyo/tinygltf/blob/master/tests/tester.cc#L416

cwbhhjl:
@syoyo Thanks! Maybe I should also report this api issue to Microsoft.

syoyo:
@cwbhhjl I think it won't be an API issue. You first convert utf8 string to MBCS string and then call `ExpandEnvironmentStringsA` on codepage 936 would give an expected result.

PR #278, opened 2020-07-10
Use std::string for GetKey function

### LINKED ISSUE #277: Invalid names retrieved from GetKey function
**Describe the issue**

When using recent versions of tinygltf on iOS, I immediately get an error in my code. This is because `primitive.attributes` has garbage in its keys. 

**To Reproduce**

- OS: iOS
- Xcode 11.5/clang, NOT using `TINYGLTF_USE_RAPIDJSON`

**Expected behaviour**

The keys in the `primitive.attributes` array match the ones from earlier.

**Additional context**

I've diagnosed the issue to the new `GetKey` function made for rapidjson: https://github.com/syoyo/tinygltf/blob/1739d025a9a833b00365ab3990df3f0cf974f572/tiny_gltf.h#L3003

When compiling with `nlohmann_json` the second line of code is run. However, the `key()` there is a temporary so returning a `char*` pointer to its `c_str()` is invalid.

**Possible solutions**

Return `std::string` instead. If that is too expensive, return a `const char*` for `rapidjson` and an `std::string` for `nlohmann_json`.

rbsheth:
Fixes #277 as seen on iOS.

Please take a look @syoyo

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -1253,7 +1253,8 @@ struct FsCallbacks {
 bool FileExists(const std::string &abs_filename, void *);
 
 ///
-/// Expand file path(e.g. `~` to home directory on posix, `%APPDATA%` to `C:\Users\tinygltf\AppData`)
+/// Expand file path(e.g. `~` to home directory on posix, `%APPDATA%` to
+/// `C:\Users\tinygltf\AppData`)
 ///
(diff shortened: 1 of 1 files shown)


syoyo:
Super awesome! Merged!

Yes it should return `std::string` to avoid referencing(pointing) a temporal variable.

PR #290, opened 2020-09-28
Minor change to the example of usage with opengl (basic example)

Coast-Coader:
The path to the model was incorrect.
With intel graphics HD (onboard on many laptops) there is a segfault happening. This can be overcome by enabling experimental mode of glew. I added the corresponding comment and the line under the glwindowhints.


syoyo:
> This can be overcome by enabling experimental mode of glew.

Thanks!

> The path to the model was incorrect.

This depends on how your working directory is set.

PR #296, opened 2020-10-29
TINYGLTF_PRESERVE_IMAGE_CHANNELS

4DA:
It might be useful to allow user to load image as is, without
complementing image channels to four.

One notable example is using both occlusion-metallic-roughness and pure
occlusion textures in project: user might use component count (1 vs 3)
to determine whether texture encodes for pure occlusion or OMR.

To do so, out should enable `TINYGLTF_PRESERVE_IMAGE_CHANNELS` compile option.


syoyo:
Awesome!

I think its better to add API to set the flag into TinyGLTF class instead of compile option: e.g.

```
void TinyGLTF::SetPreserveImageChannels(bool onoff);
```

@4DA How do you think?

4DA:
All that happens inside `LoadImageData` free function, which is **outside** from TinyGLTF class. 

I can make the value and setters/getters static but it doesn't seems right.

syoyo:
@4DA I made a PR to implement preserve image channels feature through adding new API. Could you please try this PR works well for you? 

https://github.com/syoyo/tinygltf/pull/297

4DA:
@syoyo works perfectly fine, thank you!

syoyo:
@4DA Thanks! PR #297 has been merged so close this PR also.

PR #303, opened 2021-01-21
Issue #301. Serialize accessor min/vax value as int

### LINKED ISSUE #301: Accessors min/max are serialized as float for int accessors
**Describe the issue**

All accessors min/max are serialized as double/float even it accessor refers to int values.
It raises the following validation warning:
```
{'code': 'INTEGER_WRITTEN_AS_FLOAT',
 'message': 'Integer value is written with fractional '
 'part: 25.0.',
 'pointer': '/accessors/149/max',
 'severity': 1},
```
(I might be missing something, sorry if it's the case).

**To Reproduce**

Create a glTF with accessors pointing to int values(indices for example), specify min/max, and then serialize.

**Expected behaviour**

Accessor type (int/float) is respected when serializing min/max values.

**Additional context**
It is "not an issue" when using the version with picojson since it seemed to convert `5.0` to `5` but It's not clear if the opposite was happening sometimes.
It seems to happen here in [json.hpp](https://github.com/syoyo/tinygltf/blob/master/json.hpp#L11488-L11543). 

Also these fields are optional (except for [POSITION](https://github.com/syoyo/tinygltf/blob/master/json.hpp#L11488-L11543) and [animation inputs](https://github.com/syoyo/tinygltf/blob/master/json.hpp#L11488-L11543)) so it's not a big deal, and it's possible to workaround by only serializing min/max for float accessors.

syoyo:
This will fix #301 

@AurL Could you please conform this PR fixes your issue?

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -27,7 +27,8 @@
 
 // Version:
 //  - v2.5.0 Add SetPreserveImageChannels() option to load image data as is.
-//  - v2.4.3 Fix null object output when when material has all default parameters.
+//  - v2.4.3 Fix null object output when when material has all default
+//  parameters.
 //  - v2.4.2 Decode percent-encoded URI.
 //  - v2.4.1 Fix some glTF object class does not have `extensions` and/or
 //  `extras` property.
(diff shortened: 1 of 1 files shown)


AurL:
I just did a quick test and it's ok now, the validator is happy 👍 
Thanks a lot for the quick fix!

syoyo:
@AurL Thanks for the test! Merged!

PR #305, opened 2021-02-01
check correct `extensionsUsed` for serialization

Selmar:
Check the size of the correct `extensionsUsed` variable before serialization.


syoyo:
@Selmar Super awesome! Thanks!

PR #307, opened 2021-02-26
Add example for creating a triangle gltf file from scratch

dov:
This example doesn't contain any make files, but it should be trivial to add.


syoyo:
@dov Thanks! I have added Makefile and CMake file and merged!

PR #314, opened 2021-03-20
Add (void)err for handling unused variable warning

Snowapril:
I saw in the `ParseExtensionsProperty`(3667) function,
unused variable err argument is wrapped with (void) keword.

I think err argument in the `ParseDracoExtension`(4472) is exactly
same situation with the above one.

For consistency, I think it need to be added (void)err in the
`ParseDracoExtension` function.


syoyo:
Good catch!

We need to fill  `err` message property(both for `ParseDracoExtensions` and `ParseExtentensionsProperty`) when there is an error. This is TODO.

PR #317, opened 2021-03-29
Delete unused variable initialization code

Snowapril:
Sorry for similar PR continuously.

Line 4536 in tinygltf.h file, there is unused variable initialization code like below.
```c++
    const auto pBuffer = pAttribute->buffer();
```

gcc compiler complained about that unused variable. 

I found there is **no side-effect** in  `pAttribute->buffer()` call and also pBuffer variable is **never used**. Therefore I delete that line entirely.


syoyo:
@Snowapril Super awesome! :tada: Thanks! :pray:

PR #325, opened 2021-05-11
Store scene extras and extensions on the scene

iceydee:
This was being stored in the wrong place


syoyo:
@iceydee Thanks! Good find! Merged!

PR #326, opened 2021-05-14
Fix comment on minFilter

xzrunner:


syoyo:
@xzrunner Good find! Thanks!

PR #327, opened 2021-05-20
update enum Type

CppProgrammer23:
the first element in an enum is 0, so it's not necessarily to affect values from 0 to 7.


syoyo:
@CppProgrammer23 Awesome! Thanks! Merged!

NOTE: https://stackoverflow.com/questions/34811486/do-c-enums-start-at-0 In C++, its guaranteed to have zero for the first element when no initializes, whereas C is not.

PR #331, opened 2021-10-14
Remove stackoverflow references from tinygltf.

abwood:
In very few places in this header file (I count 4), there are code snippets that reference stackoverflow. The licensing of all content posted on stackoverflow is [CC BY-SA](https://creativecommons.org/licenses/by-sa/4.0/) ([source](https://stackoverflow.com/help/licensing)). The use of a ShareAlike clause makes this a copy-left license, implying that use of this code in any software requires that it also be released under the same license as the original (as best as I understand this). As this seems incompatible with tinygltf's MIT license, it would be best to review and remove these references.

I have made the following changes that I hope will not only separate tinygltf from stackoverflow, but also provide some (minor) improvements to the library.

The changes are as follows:
- [x]  GetBaseFilename now correctly handles the case where a filename is provided that contains no path separator characters
- [x]  Removed the roundUp lambda, as all call-sites used this to round up to 4 byte alignment; I just calculated the necessary padding for 4-byte alignment for the two buffers that we may serialize.
- [x] Remove clang warning suppression. I assume this stackoverflow reference can just be removed as it isn't a code snippet but a repost of clang documentation of ignoring warning levels.
- [x] Remove stackoverflow reference that recommends the use of dlib. There's no source that was pulled from this s.o. post, only a recommendation on a library, I wouldn't think a link to s.o. is necessary here. Per the Boost License, I've included a copy of the license along with the snippet.

~~Tests added to verify round trip model write and model read to verify code changes.~~ Manual testing by @syoyo.


syoyo:
@abwood Super awesome! Let me give some time to review PR and do testing that changes does not break (unit) tests.

syoyo:
@abwood I ran some local tests on my side, and everything works fine.

Do you have a plan to add regression test for this?

> Tests added to verify round trip model write and model read to verify code changes.

If you have a plan to add regression tests in short-term(~a couple of weeks), I'd like to wait for it. If not, I'll merge PR anyway(and may add regression tests later)

PR #339, opened 2021-11-28
update json.hpp to v3.10.4 to compile with c++20 compilers

yslking:
When I use tinygltf with a C++ 20 compiler it complains that std::is_pod (https://en.cppreference.com/w/cpp/types/is_pod) in json.hpp has been deprecated since C++ 20 . The upstream has replaced std::is_pod with std::is_trivial and std::is_standard_layout (nlohmann/json@61b46a5) . Could we upgrade json.hpp to the latest release version (which is v3.10.4) ?

If the team thinks its better to retain current version of json.hpp, then could we at least replaced std::is_pod with std::is_trivial and std::is_standard_layout ( as what the upstream did nlohmann/json@61b46a5) to satisfy C++ 20 compilers


syoyo:
@yslking Thanks! Merged! BTW you can include your own `json.hpp` with `TINYGLTF_NO_INCLUDE_JSON` compile flag: https://github.com/syoyo/tinygltf#compile-options

PR #342, opened 2022-02-23
add missing switch case for binary

brlcad:
with warnings cranked up and warnings-as-errors enabled, the compiler throws an error due to this switch missing a case for binary.  this makes the ignore explicit in order for compilation to proceed.


syoyo:
@brlcad Thanks! Merged!(There are some build failures, but these are  CI issues so we can ignore it)

brlcad:
@syoyo sure thing!  Just FYI, we’re using your library in BRL-CAD for out gltf import support (and may use it for export support).  It’s an open source solid modeling system that has been around for decades, specializes in analysis-related CAD uses.  Thanks!

PR #344, opened 2022-02-24
Added missing comparison for wrapS

### LINKED ISSUE #343: Typo in Sampler::operator==  ?
**Describe the issue**

I decided to update my tinygltf version and while comparing the differences I noticed that
wrapS is not checked when doing Sampler::operator==

The current code is as follows:

```
bool Sampler::operator==(const Sampler &other) const {
  return this->extensions == other.extensions && this->extras == other.extras &&
         this->magFilter == other.magFilter &&
         this->minFilter == other.minFilter && this->name == other.name &&
         this->wrapT == other.wrapT;

         //this->wrapR == other.wrapR && this->wrapS == other.wrapS &&
}
```
Am I missing something or was the wrapS removed incorrectly when wrapR was removed?

**To Reproduce**

Issue noticed by looking at source code

**Expected behaviour**

wrapS should also be checked

**Screenshots**

no screenshots

**Additional context**

epajarre:
Fixing missing wrapS comparison, it was removed by mistake when wrapR was removed
Fixes #343

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -1933,9 +1933,10 @@ bool Sampler::operator==(const Sampler &other) const {
   return this->extensions == other.extensions && this->extras == other.extras &&
          this->magFilter == other.magFilter &&
          this->minFilter == other.minFilter && this->name == other.name &&
+         this->wrapS == other.wrapS &&
          this->wrapT == other.wrapT;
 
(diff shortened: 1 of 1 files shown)


syoyo:
@epajarre Thanks! Merged! There is a Windows CI setup issue, but we can ignore it for now.

PR #346, opened 2022-02-24
Fix byte offset optionality for sparse accessor indices and values

jmousseau:
[`accessor.sparse.indices.byteOffset`](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/accessor.sparse.indices.schema.json#L45) and [`accessor.sparse.values.byteOffset`](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/accessor.sparse.values.schema.json#L22) are both optional with default values of zero. I tested with [`Accessor_Sparse_01.gltf`](https://github.com/KhronosGroup/glTF-Asset-Generator/blob/master/Output/Positive/Accessor_Sparse/Accessor_Sparse_01.gltf#L103-L112), which contains a sparse accessor without a `byteOffset` key for its `indices` and `values.`

```bash
./loader_example glTF-Asset-Generator/Output/Positive/Accessor_Sparse/Accessor_Sparse_01.gltf
```

Output before:

```txt
Reading ASCII glTF
Err: 'byteOffset' property is missing.
'byteOffset' property is missing.
```

Output after:

```txt
Reading ASCII glTF
```

The sample model test runner and unit tests still succeed.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -4238,7 +4238,7 @@ static bool ParseSparseAccessor(Accessor *accessor, std::string *err,
   ParseIntegerProperty(&indices_buffer_view, err, indices_obj, "bufferView",
                        true);
   ParseIntegerProperty(&indices_byte_offset, err, indices_obj, "byteOffset",
-                       true);
+                       false);
   ParseIntegerProperty(&component_type, err, indices_obj, "componentType",
                        true);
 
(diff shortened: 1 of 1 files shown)


syoyo:
@jmousseau Good find! Thanks!

> The sample model test runner and unit tests still succeed.

Let me give some time to review why current unit test passes before the merge

syoyo:
https://github.com/syoyo/tinygltf/blob/e59028d68f7116f38385e4c04b6d3493313b4df9/tiny_gltf.h#L4258

Ah, error checking for SparseAccessor properties are missing...

syoyo:
@jmousseau  Added additional checks over your PR and confirmed unit test passes. > https://github.com/syoyo/tinygltf/pull/347

Thank you for PR!

PR #349, opened 2022-03-01
Fix examples/glview rotations

tsmaster:
GLtf defines rotations as quaternions, glRotated expects angle/axis, with angle in degrees.


syoyo:
@tsmaster Super awesome! Thanks! Merged!

tsmaster:
In additional testing (with the other cars in Kenney's Car Kit), I think
the Translation/Rotation/Scale matrices have the scale in the wrong order.
I'll send another PR for that soon.

PR #350, opened 2022-03-02
Examples/glview - apply translation, rotation, scale in that order

tsmaster:
Tested on the Kenney.nl CarKit models: https://www.kenney.nl/assets/car-kit


syoyo:
@tsmaster Thanks! Merged! (Yes, `T x R x S` order https://github.com/KhronosGroup/glTF/issues/743 )

PR #353, opened 2022-03-13
Fixed find_last_of call in GetBaseFilename

mgerhardy:
this leads to a stack overflow on linux - it looks like this magically works on msvc though...

Already in use in vengi https://github.com/mgerhardy/vengi/commit/2c8716fbbe8be30eda99944c5291d373cd6610d1

https://www.cplusplus.com/reference/string/string/find_last_of/


syoyo:
@mgerhardy Oh Good find! Merged! Thanks!

PR #357, opened 2022-05-14
Fix basic example

eariassoto:
Context: When parsing a mesh with primitives that have indices pointing to different buffer views, the VAO will use the last GL_ELEMENT_ARRAY_BUFFER that was bound to it. In this scenario, the mesh will not be rendered properly.

When binding a mesh, return VAO and EBOs so that primitives can bind to the proper element array buffer.

### DIFF
--- a/examples/basic/main.cpp
+++ b/examples/basic/main.cpp
@@ -39,8 +39,8 @@ bool loadModel(tinygltf::Model &model, const char *filename) {
   return res;
 }
 
-std::map<int, GLuint> bindMesh(std::map<int, GLuint> vbos,
-                               tinygltf::Model &model, tinygltf::Mesh &mesh) {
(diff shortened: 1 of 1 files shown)


syoyo:
> When parsing a mesh with primitives that have indices pointing to different buffer views

@eariassoto Awesome! Is it possible to share such glTF data for a regression test on my side?

eariassoto:
Sure thing, this is a modified version of the Cube.gltf that has 2 primitives, each with half of the cube triangles: https://gist.github.com/eariassoto/cca35c25ac7b02bafa718557518b13cc
I did not modified the original, so as to not introduce potential regressions in the other examples.

syoyo:
Thanks! Confirmed the issue with given .gltf  >  the mesh will not be rendered properly.

And PR has been merged!

Issue #359, opened 2022-05-22
Make writing GLTF scene to a stream more flexible


TheMostDiligent:
**Describe the issue**

Currently, `WriteGltfSceneToStream` always embeds images and buffers into the output stream, which is not very flexible. In our usage scenario we would like to write images to separate files as they are shared between multiple scenes, but at the same time to be able to output the scene to a custom stream. At the moment we have to write the scene to a file on disk and then immediately read it back.

**To Reproduce**

- OS: any
- Compiler, compiler version, compile options: any

**Expected behaviour**

`WriteGltfSceneToStream` takes additional arguments `embedImages` and `embedBuffers` that control how images and buffers are embedded. The following function overload could be added which will let the existing function work as before:

```cpp
  bool WriteGltfSceneToStream(Model *model, std::ostream &stream,
                              const std::string &base_dir,
                              bool embedImages, bool embedBuffers,
                              bool prettyPrint, bool writeBinary);
```

**Screenshots**

n/a

**Additional context**

n/a

syoyo:
It may be better to introduce callback handlers to `WriteGltfSceneToStream` for customizing Image output and Buffer output.

Your contribution is much appreciated!

TheMostDiligent:
There are already file system callbacks that can be set with `SetFsCallbacks`. And as a matter of fact this is how I tried to implement it initially - by proving a custom `WriteWholeFile` function. I was expecting that the callback will be called for the GLTF file.
It however turned out that `WriteWholeFile` is only used to write images when `TINYGLTF_NO_STB_IMAGE_WRITE` is not defined and is never used otherwise.
WriteGltfFile always uses `std::ofstream` for writing the file.

syoyo:
> It however turned out that WriteWholeFile is only used to write images when TINYGLTF_NO_STB_IMAGE_WRITE is not defined and is never used otherwise.

Oh, I see.. Writing API is not mature and needs an improvement.

Anyway, You can propose PR to fix/improve writing API!

TheMostDiligent:
How would you suggest to implement this - to make the callbacks work or use the new version of `WriteGltfSceneToStream` function?

TheMostDiligent:
[...]

PR #362, opened 2022-06-29
fix images not beeing embeded into stream

hschwane:
The documentation of `WriteGltfSceneToStream` states that `buffers and images will be embeded`. When testing I had quite some trouble to teach the exporter to include my images (I am building the gltf (including images) from scratch for exporting). I noticed `embedImages` was set to false when calling `WriteGltfSceneToStream`. I assume this is a bug, so this PR fixes that.


syoyo:
Thanks! Merged!

TODO: Make `WriteGltfSceneToStream` API more flexible https://github.com/syoyo/tinygltf/issues/359

PR #366, opened 2022-08-13
Adding serialization code for sparse accessors.

fynv:
Tested in my FBX->GLB converter. 

Gestaltor seems to recognize the converted sparse morph targets correctly:

image


syoyo:
@fynv Super awesome! Thanks! Merged!

PR #367, opened 2022-08-14
Sparse accessor: typo fix

fynv:
Sorry I made a mistake with accessor.sparse.values.byteOffset.

The value is not used at all in some applications, so the test didn't detect it.


syoyo:
Good catch! Thanks! Merged!

Issue #372, opened 2022-08-25
Heap buffer overflow when parsing glb binary file


Pyraun:
**Description**

The [`TinyGLTF::LoadBinaryFromMemory`](https://github.com/syoyo/tinygltf/blob/master/tiny_gltf.h#L6267) function does not check the size of the `.glb` input file properly.
Any program trying to parse a malicious binary `.glb` file through a call of the `TinyGLTF::LoadBinaryFromMemory` will most likely crash, which could result in an Denial Of Service.

The binary gltf format has a 20 bytes long header which contains two fields: `length` and `model_length`.
These two fields are unsigned integer of 4 bytes (`unsigned int`).

```cpp
  unsigned int version;       // 4 bytes
  unsigned int length;        // 4 bytes
  unsigned int model_length;  // 4 bytes
  unsigned int model_format;  // 4 bytes;

  /* ... */

  // In case the Bin buffer is not present, the size is exactly 20 + size of
  // JSON contents,
  // so use "greater than" operator.
  if ((20 + model_length > size) || (model_length < 1) || (length > size) ||
      (20 + model_length > length) ||
      (model_format != 0x4E4F534A)) {  // 0x4E4F534A = JSON format.
    if (err) {
      (*err) = "Invalid glTF binary.";
    }
    return false;
  }

  // Extract JSON string.
  std::string jsonString(reinterpret_cast<const char *>(&bytes[20]),
                         model_length);
```

The checks on `model_length`: `model_length < 1`, `20 + model_length > size` and `20 + model_length > length` fail when an integer overflow occurs.
If `model_length = 0xffffffed`, the expression `20 + model_length` equals `1`. In this case every check succeeds and yet the `model_length` used to instantiate the `jsonString` is a value way greater than the size of the `bytes` buffer.
The creation of the `std::string` object ends up into a call to the `memcpy` function with a copy size equals to `model_length = 0xffffffed`, which results in a crash of the process.

**Proof Of Concept**

The following python script (`gen_glb.py`) can used to generate a `.glb` file that will crash the [`loader_example`](https://github.com/syoyo/tinygltf/blob/master/loader_example.cc) program present in the tinygltf repository.

- a testcase is also available here: [poc.zip](https://github.com/syoyo/tinygltf/files/9427345/poc.zip)

```python
from pathlib import Path
from struct import pack

import argparse

parser = argparse.ArgumentParser()
parser.add_argument("-o", "--output", default=Path("test.glb"))
[...]

PR #374, opened 2022-08-26
Better GLB data size check when reading.

syoyo:
GLB: Check BIN Chunk size.
GLB: Chenk byte alignment of each chunk.
GLB: Fix integer-overflow when calculating JSON Chunk size(fixes https://github.com/syoyo/tinygltf/issues/372 )


Pyraun:
@syoyo 

It sounds good to me and a small fuzzing campaign with libfuzzer and the regression test as corpus did not found other issues.

syoyo:
@Pyraun :pray:

PR #376, opened 2022-09-02
Read from moved object

kacprzak:
f4f5c3c - We should not read from object that was moved in previous line.
b12a54e - Second commit is just silencing warning. But also for me at firs read it looked like a mistake.


syoyo:
Awesome! Thanks! Merged!

> We should not read from object that was moved in previous line.

Good catch!

> But also for me at firs read it looked like a mistake

Yes probably something like a mistake...

PR #377, opened 2022-09-05
Auto detect C++14 standard version

zbendefy:
This should auto-detect if tinygltf is being compiled against c++14 or newer standard, to enable the optimizations near the end of tiny_gltf.h. Note that MSVC requires '/Zc:__cplusplus' compiler flag so that __cplusplus is set correctly (https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/)

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -66,6 +66,11 @@
 #include <string>
 #include <vector>
 
+//Auto-detect C++14 standard version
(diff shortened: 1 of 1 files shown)


syoyo:
@zbendefy Super awesome! Thanks! Merged!

PR #379, opened 2022-09-06
Fix possible out of bounds index in LoadFromString

nirmal:
**Describe the issue**

We triggered a crash in TinyGLTF::LoadFromString by passing in a bad input. The accompanying PR fixes these issues and LoadFromString returns false as expected.

**To Reproduce**

[invalid_mesh.gltf.txt](https://github.com/syoyo/tinygltf/files/9498851/invalid_mesh.gltf.txt) - pass to TinyGLTF::LoadFromString to trigger the crash.

**Expected behaviour**

We expect TinyGLTF::LoadFromString to return false.

**Screenshots**

None

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -5803,20 +5803,27 @@ bool TinyGLTF::LoadFromString(Model *model, std::string *err, std::string *warn,
       }
 
       for (auto &attribute : primitive.attributes) {
-        model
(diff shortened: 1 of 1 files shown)


syoyo:
@nirmal Good catch! Thanks! Merged!

TODO: Report an warning when an index is out-of-range.

PR #380, opened 2022-09-10
Fixes compiler warning on VS (Unreachable code detected)

AlvaroBarua:
I use warnings as errors and this wouldn't let me compile, I doubt I am the only one who has encountered this.


syoyo:
@AlvaroBarua Thanks! Mereged!

PR #382, opened 2022-09-17
Fix parsing GLB file with empty Chunk1(BIN data).

syoyo:
This PR does not check Chunk1(BIN) size when its empty(omitted) from GLB, to load GLB with empty BIN buffer correctly.

The issue was reported here
https://github.com/syoyo/tinygltf/pull/382#issue-1376739827


Kh4n:
Wow you are fast! I have tested this with the following results:
empty gltf: pass
empty gltf with single empty buffer: fail
empty gltf with buffer that has one byte: fail

This works for me, I believe the other cases are correct as well. I have created tests for them in the tester.cc file, I can add a PR for that, however I noticed that the gitignore is set to ignore lots of files in the repo that are being committed, should we do anything about that? It made a bit of a mess after `git rm -rf --cached .` :) (i did that to remove some test files that were generated that I committed by accident)

syoyo:
@Kh4n Thanks!

- empty gltf with single empty buffer: fail
  => this may be ok to fail. glb says chunk1 should be completely omitted when bin data is empty: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#binary-buffer 
- empty gltf with buffer that has one byte: fail
  => this may be passed. I modified to reject buffer whose size is less than 4 bytes

As you may know, you can validate .glb file using gltf-validator: https://github.khronos.org/glTF-Validator/

`.gitignore` should not disturb creating PR if your changes are composed of  `*.cc` and `*.glb` files. If you still struggle to create PR, alternatively you can attach `.glb` testcases files here. I'll investigate it and add it as a regression testcases in my side.

Kh4n:
https://github.com/syoyo/tinygltf/pull/383 added into this PR (the test that is)

syoyo:
@Kh4n Thanks! Will look into PR #383 💪

Kh4n:
Based on #383 I think this is ready to go

Kh4n:
Looks good, just one small mistake in my test, there should be an `err.clear()` right after the second test ([right here for example](https://github.com/syoyo/tinygltf/pull/382/files#diff-c32c0f21707473e9ca994e600c04ec16d971f14e9dacc6e5cd6373a70fcdfeedR516)), that way the error does not get printed which could be confusing, besides that everything looks good and ready to go

syoyo:
@Kh4n  

> there should be an err.clear() right after the second test

I see. I've added clear() for err(and also warn just in case). And merged! Thank you for the contribution!

PR #388, opened 2022-10-30
Avoid multiple asset_manager definitions on Android

stromaster:
Using TINYGLTF_IMPLEMENTATION to make the actual definition of the 'asset_manager' global variable.


syoyo:
@stromaster Super awesome! Thanks!

PR #390, opened 2022-11-07
Added Supernova Engine to Projects using TinyGLTF

eduardodoria:
Supernova is a free and open source cross-platform game engine for create 2D and 3D projects with Lua or C++. It is lightweight and promote the simplest way to use de best performance of data-oriented design. It uses tinygltf.


syoyo:
@eduardodoria Awesome! Merged!

PR #393, opened 2022-12-21
allow serializing a const Model

pknowlesnv:
Adds 'const' to all Serialize*() methods.

Updates WriteImageData callback to take a URI out pointer that was previously being written to the Image::uri, which is now const.

This breaks the WriteImageData API and as a side effect, Image::uri will no longer contain the written image URI after saving.

### DIFF
--- a/tests/tester.cc
+++ b/tests/tester.cc
@@ -540,3 +540,66 @@ TEST_CASE("empty-bin-buffer", "[issue-382]") {
   }
   REQUIRE(true == ret);
 }
+
+TEST_CASE("serialize-const-image", "[issue-394]") {
+  tinygltf::Model m;
+  tinygltf::Image i;
+  i.width = 1;
+  i.height = 1;
+  i.component = 4;
+  i.bits = 8;
+  i.pixel_type = TINYGLTF_COMPONENT_TYPE_UNSIGNED_BYTE;
+  i.image = {255, 255, 255, 255};
+  i.uri = "image.png";
+  m.images.push_back(i);
+
+  std::stringstream os;
+
+  tinygltf::TinyGLTF ctx;
+  ctx.WriteGltfSceneToStream(const_cast<const tinygltf::Model *>(&m), os, false,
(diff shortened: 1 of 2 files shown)


syoyo:
@pknowlesnv `SerializeGltfImage` does not utilize `uri` argument. I guess your PR want to use `uri`, not `image.uri` ?

https://github.com/pknowlesnv/tinygltf/blob/a90a97b9a403a5175d351d2d84f75df969517010/tiny_gltf.h#L7013

pknowlesnv:
Aah right. Updated. Thanks!

What are your thoughts re. adding the `out_uri` parameter to `WriteImageDataFunction`?

While testing I made these follow-up changes too. Let me know if either look worth keeping.

https://github.com/pknowlesnv/tinygltf/pull/1

https://github.com/pknowlesnv/tinygltf/pull/2

syoyo:
@pknowlesnv Thanks! 

> What are your thoughts re. adding the out_uri parameter to WriteImageDataFunction?

Yes I definitely like it! Although it breaks current Callback API signature, people who are currently using serialization feature should easily migrate new API. 

Do you have a glTF file(minimal one preferred) or (unit) test codes to test whether `out_uri` is actually used in serialized(produced) glTF file?

> While testing I made these follow-up changes too.

These are also awesome!

pknowlesnv:
@syoyo Good call on the tests. I have some code locally, but I see there's a `tester.cc` file which is perfect. I added a couple of cases.

syoyo:
@pknowlesnv Thanks! Confirmed unit test passed on my side, so merged!


Also, it'd be super awesome if you send follow-up changes https://github.com/syoyo/tinygltf/pull/393#issuecomment-1366290061 as another PRs!

pknowlesnv:
Thanks to you for making and maintaining all these awesome tools :)

Created https://github.com/syoyo/tinygltf/pull/396 and https://github.com/syoyo/tinygltf/pull/397

PR #395, opened 2022-12-28
Add CMake export configuration

ethanpepro:
I needed an import target for TinyGLTF as a convenience upgrade in my personal projects. This defines the `tinygltf::tinygltf` target for CMake projects and automatically includes the headers and the optional static library. This is a minimal but functional change so as not to break any existing install features and is written to match the same indentation style as the existing `CMakeLists.txt`.

I currently use it in my project as such:
`find_package(tinygltf HINTS ${CMAKE_PREFIX_PATH}/tinygltf PATH_SUFFIXES lib/cmake REQUIRED)`
`target_link_libraries(... PRIVATE ... tinygltf::tinygltf)`


syoyo:
@ethanpepro Thanks! Merged!

> I currently use it in my project as such:

TODO: Write this `find_package(tinygltf)`  cmake script code fragment in README.md

PR #396, opened 2022-12-29
propagate image writing failures

pknowlesnv:
Modifies UpdateImageObject() so that Returning false from the WriteImageDataFunction callback results in the WriteGltfScene*() call returning false.

Does not call WriteImageDataFunction if there is no image data.

Adds test case serialize-image-failure to verify the callback return code is able to cause an overall failure to save the gltf.

(patch created while revising https://github.com/syoyo/tinygltf/pull/393#issuecomment-1366290061)

### DIFF
--- a/tests/tester.cc
+++ b/tests/tester.cc
@@ -593,13 +593,38 @@ TEST_CASE("serialize-image-callback", "[issue-394]") {
 
   tinygltf::TinyGLTF ctx;
   ctx.SetImageWriter(writer, (void *)0xba5e1e55);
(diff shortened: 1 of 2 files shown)


pknowlesnv:
I see there's a [devel](https://github.com/syoyo/tinygltf/tree/devel) branch too. Let me know if there's a convention for contributing.

syoyo:
@pknowlesnv Thanks! Will check it soon!

syoyo:
@pknowlesnv Merged! Thank you for the contribution!

PR #399, opened 2023-01-13
Missed serialization texture sampler name fixed.

e2e4e6:
According to the glTF 2.0 specification it exists, the internal tinygltf structure contains 'name' field as well.


e2e4e6:
Links to the specification:
https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_sampler_name
https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/sampler.schema.json

syoyo:
Thanks! Merged!

A build failure is due to Azure error, so we can ignore it.

PR #401, opened 2023-01-15
Fix animation channel target node optionality

jmousseau:
[`animation.channel.target.node`](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/animation.channel.target.schema.json#L36) is an optional value. For example, in the [`KHR_animation_pointer`](https://github.com/KhronosGroup/glTF/blob/95e28d582382d9564d4ccc5706ae4640d2a14377/extensions/2.0/Khronos/KHR_animation_pointer/README.md) extension, a new `"pointer"` path is introduced that indicates the JSON pointer in the extension should be used instead of a `node` target.

Tested with the sample animation pointer asset provided in https://github.com/KhronosGroup/glTF-Sample-Models/pull/353.

```bash
./loader_example glTF-Sample-Models/2.0/AnimateAllTheThings/glTF/AnimateAllTheThings.gltf
```

Output before:
```txt
Reading ASCII glTF
Err: 'node' property is missing.
`node` field is missing in animation.channels.target
```

Output after:
```txt
Reading ASCII glTF
```

The sample model test runner and unit tests still succeed. To keep the changes minimal, I just fixed optionality instead of adding full `KHR_animation_pointer` parsing support.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -547,9 +547,10 @@ typedef std::map<std::string, Value> ExtensionMap;
 
 struct AnimationChannel {
   int sampler;              // required
-  int target_node;          // required (index of the node to target)
-  std::string target_path;  // required in ["translation", "rotation", "scale",
-                            // "weights"]
+  int target_node;          // optional index of the node to target (alternative
+                            // target should be provided by extension)
+  std::string target_path;  // required with standard values of ["translation",
+                            // "rotation", "scale", "weights"]
   Value extras;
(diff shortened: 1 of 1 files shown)


syoyo:
@jmousseau Good catch! Thanks! 

Probably I wrote the corresponding code when `node` was a requited item 🤔 https://github.com/KhronosGroup/glTF/commit/3d3db0455790d510e23505cd30da270c92502f43

Issue #406, opened 2023-02-17
Consider using `nlohmann::json` instead of just `json`


DavidSM64:
**Describe the issue**

In my project I have been using the SimpleJSON library (https://github.com/nbsdx/SimpleJSON) for creating/loading json files, which uses the namespace `json` for its classes. When trying to use the tinygltf library I get ambiguity errors because you are also using `json`. I solved this issue in my own project by changing `json` to `nlohmann::json` in tiny_gltf.h

**To Reproduce**

- OS: Ubuntu (Under WSL2)
- Compiler: g++

**Expected behaviour**

Changing `json` to `nlohmann::json` in the nlohmann implementation fixes the problem. I know it would be annoying, but it should help resolve future conflicts.

syoyo:
Oh, I see!

 `json` namespace is not exposed to tinygltf public API, so it should be safe to rename it(for example, `json_backend`)

Also, we'd be better to wrap some implementation code with `detail`(or `internal`) namespace(instead of anonymous namespace https://github.com/syoyo/tinygltf/blob/6dba104c54bebd6c6b34f589272a06da509156e2/tiny_gltf.h#L1708 ) to avoid potential namespace conflicts issue further. i.e.

```c++

namespace tinygltf {
namespace detail {

using json_backend = nlohnann::json;

...

void JsonParse(...) {
  doc = json_backend::parse(...)
}

} // detail
} // tinygltf
```

PR is much appreciated!

syoyo:
🙏

PR #407, opened 2023-02-18
Added detail namespace to prevent json namespace conflicts

DavidSM64:
Fixes issue #406 

Ran parser & unit tests, and it all passed.

Don't even need to rename `json` in this case. For some weird reason it has to use `::detail::json` instead of `detail::json`. I have no idea why.

### DIFF
--- a/tests/tester.cc
+++ b/tests/tester.cc
@@ -16,10 +16,10 @@
 #include <sstream>
 #include <fstream>
 
-static detail::JsonDocument JsonConstruct(const char* str)
+static tinygltf::detail::JsonDocument JsonConstruct(const char* str)
 {
-  detail::JsonDocument doc;
-  detail::JsonParse(doc, str, strlen(str));
+  tinygltf::detail::JsonDocument doc;
+  tinygltf::detail::JsonParse(doc, str, strlen(str));
   return doc;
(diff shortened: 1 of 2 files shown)


syoyo:
@DavidSM64 Thanks! But it fails to compile with rapidjson backend.

Also, let me give some to resolve `::detail` namespace scope issue. We should be able to make it `details::`

DavidSM64:
RapidJSON should be fixed now hopefully. Going to look into the `::detail` issue.

Edit: Ah, found the issue. Forgot to add `namespace tinygltf {` above a `namespace detail {`. This should be good to go now if you don't have any other issues with it.

syoyo:
@DavidSM64 Thanks!

> RapidJSON should be fixed now hopefully.

Awesome!

> Forgot to add namespace tinygltf { above a namespace detail {

Oh, some JSON stuff was defined in global unnamed namespace scope and not wrapped by `tinygltf` namespace 😅 

https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces

Your PR makes it better.

syoyo:
@DavidSM64 Confirmed CI build passes, so merged! Thank you for the contribution!

PR #412, opened 2023-04-08
Fix #411 by adding a Value(const char *) constructor.

### LINKED ISSUE #411: [Defect] tinygltf::Value(char const*) invokes tinygltf::Value(bool) unexpectedly
Hi @syoyo,

I'm currently playing with your loader. So first, thanks for making this!

I came across a little issue with `tinygltf::Value`, in particular with the massively overloaded constructor...

**Describe the issue**

Constructing values manually using a `char *` as argument should result in a `tinygltf::Value` of type string. Instead it invokes the bool constructor, because of the automatic conversion of any pointer type to bool.

**To Reproduce**

- OS: Any
- Compiler: Any

```
#include <iostream>
#include <tiny_gltf.h>

int
main(int argc, char * argv[]) {
  tinygltf::Value str_val("foo");
  std::cout << "str_val type: " << (int)str_val.Type() << std::endl;
  assert(str_val.Type() == tinygltf::STRING_TYPE);
  return 0;
}
```

As a workaround `tinygltf::Value str_val(std::string("foo"));` works as expected. Interestingly, `tinygltf::Value str_val {"foo"};` fails at compile-time, because of the stricter type narrowing rules in C++11.

**Expected behaviour**

The code above should produce a value of type string (or fail at compile-time). Adding yet another constructor overload for `char const*` should do the trick... I think.

Let me know what you think...

Edit: typos

agnat:
Avoid implicit conversion of pointers to bool. Closes #411.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -301,6 +301,9 @@ class Value {
   }
   explicit Value(std::string &&s)
       : type_(STRING_TYPE), string_value_(std::move(s)) {}
+  explicit Value(const char *s) : type_(STRING_TYPE) {
+    string_value_ = s;
+  }
   explicit Value(const unsigned char *p, size_t n) : type_(BINARY_TYPE) {
     binary_value_.resize(n);
     memcpy(binary_value_.data(), p, n);


syoyo:
@agnat Thanks! There is a CI build issue. Ubuntu 18.04 image is obsolete so I'll try to update CI build env first(Although the PR should work for gcc4.8, mingw).

syoyo:
Fixed CI build issue and merged the PR! Thanks!

PR #414, opened 2023-04-15
Feature: Add entries to existing tinygltf::Values

agnat:
Hi @syoyo,
how are you doin'? Thanks for the quick merge the other day. I continued exploring and came across another small issue with `tinygltf::Value`.

Consider a typical baking job that 1. reads an asset, 2. adds some extras (or extensions) and 3. saves the file. In this scenario we'd like to add entries to an existing object incrementally without replacing the whole object, because it may already contain entries from earlier stages (eg. the exporter). I couldn't find a (good) way to do that...

This PR adds a subscript operator to insert entries into objects. To maintain feature parity I also added a push method and subscript operators for arrays.

Let me know what you think...

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -411,12 +411,12 @@ class Value {
 
   bool operator==(const tinygltf::Value &other) const;
 
-  Value & operator[](std::string const& key) {
+  Value & operator[](const std::string &key) {
     assert(type_ == OBJECT_TYPE);
     return object_value_[key];
   }
 
-  void Push(Value const& val) {
+  void Push(const Value &val) {
     assert(type_ == ARRAY_TYPE);
     array_value_.push_back(val);
   }
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -411,6 +411,29 @@ class Value {
 
   bool operator==(const tinygltf::Value &other) const;
 
+  Value & operator[](std::string const& key) {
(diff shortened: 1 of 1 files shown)


agnat:
We also could expose `Resize()` and maybe even `Reserve()` on arrays ... but I don't think that's necessary. I can add them if you want, though...

syoyo:
You are better first explicitly cast Value data to Array or Object

agnat:
Apologies, but I don't understand what you mean. Could you elaborate or paste an example?

agnat:
Hi @syoyo,
could you do me a favour and bonk the button on this? For me it's an actual issue and it would be great to get it landed.

syoyo:
`tinygltf::Value` must be rewritten to make it type safe and remove `assert` in `Value` class for better security.

For example, `ArrayLen`, and `const Value &Get(int idx) const`(and other non-type safe API) must be removed.


It'd be better to provide  type safe API to Value, such like:

```
bool GetAsArray(Value::Array *arr) {
   if (IsArray()) {
      if (arr) { (*arr) = array_value_; return true}
   }
   return false;
}  

// Do same thing for Value::Object type.
```

If you want to access array object in current Value API, you could do something like

```
if (v.IsArray()) {
  Array arr = v.Get<Value::Array>();
  for (item : arr) {
    ...
  }

  v = Value(arr); // update
}
```

PR #415, opened 2023-04-18
Fix serialization of AnimationChannel::target_node when it is zero

louwaque:
Fix serialization of AnimationChannel::target_node when it is zero.


syoyo:
Good catch! Thanks! Merged!

PR #418, opened 2023-04-26
[Bugfix] Actually invoke the user-supplied function instead of subtracting from a pointer...

agnat:
Hi @syoyo,

I spotted this bug in the new file size handling code. It's kind of funny that this is still a valid expression, although it is clearly missing a character. That's pretty rare... anyway...

Enjoy your day...


syoyo:
Good catch!  Thanks!

PR #422, opened 2023-06-05
Add missing extras and extensions fields

agnat:
Hi syoyo,

this PR adds a number of missing extras and extension fields, mostly to nested "sub-objects". I also refactored the parsing and serialization code for these fields, so it's easier to maintain.

Last, I got rid of `std::function` by rewriting `ForEachInArray(...)` as a plain old template function. Also, fixed a warning introduced in #417...

Let me know what you think...

PR #424, opened 2023-06-06
get rid of std::function

agnat:
No need for std::function here. A free function with an unspecified callback will do nicely.

This was first proposed in #422


syoyo:
Thanks! Merged!

PR #427, opened 2023-06-09
Refactor toplevel section parsing and remove duplicate code

agnat:
Introduce a new function ParseSection(…), which handles populating the toplevel GLTF arrays. This removes a bunch of duplicated code. Use a variadic template to facilitate passing additional arguments, that don’t fit the common signature. See Buffer or Mesh.

Refactor object parsing functions (ParseBuffer(), &c.) into a set of overloaded functions ParseObject(…). That way we can invoke them without knowning the actual type.

Pass the TinyGLTF context to the parser functions, so we can actually use it.

To keep the changeset size in check, this deals with all toplevel objects but Image, which is more involved and requires a closer look.

Also, add some getters and fix minor const-correctness issues.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -4329,13 +4329,6 @@ static bool ParseOcclusionTextureInfo(
 
 static bool ParseObject(Buffer *buffer, std::string *err, const TinyGLTF & ctx,
                         const detail::json &o, const std::string & basedir) {
-                        /*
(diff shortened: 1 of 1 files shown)


syoyo:
I don't like variadic template, which makes C++ code difficult to read and debug, and code refactoring is a trivial thing, so won't merge this PR

PR #429, opened 2023-06-12
KHR_audio parsing support (ASCII only)

AdamFull:
Added parsing for the official KHR_audio extension. Currently only for ASCII version (didn't add ability to read and write audio from/to glb. But I added bufferView and mimeType fields)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ premake5.tar.gz
 #binary directories
 bin/
 obj/
+out/
 
 #runtime gui config
 imgui.ini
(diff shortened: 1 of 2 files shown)


syoyo:
Awesome! Let me give some time to review this PR.

AdamFull:
I used same extension revision, that was used in gestaltor. This extension is still have status testing, but gestaltor already use it.

syoyo:
Thanks! Merged!

https://github.com/KhronosGroup/glTF/pull/2137

`KHR_audio` extension spec is not yet finalized, so treat this ext as an experimental.

TODO: Support loading some audio data(mp3, wav, etc) using dr_libs https://github.com/mackron/dr_libs

PR #436, opened 2023-07-10
Fix issue when serializing extensions of a primitive

dimitri-tdg:
Hi,

It seems a small regression has been introduced during the refactor of extensions and extra serialization. 
Note that this bug was already fixed in #230.

Regards


syoyo:
Thanks! Merged!

TODO: It'd be nice to have a regression test...

PR #445, opened 2023-08-24
Be tolerant when encountering emissiveFactor with array length 4

nyalldawson:
Raise a warning when encountering emissiveFactor with array length of 4 and truncate to the first 3 elements, instead of aborting the model loading entirely.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -195,6 +195,11 @@ typedef enum {
   OBJECT_TYPE
 } Type;
 
+typedef enum {
+  PERMISSIVE,
+  STRICT
+} ParseStrictness;
+
 static inline int32_t GetComponentSizeInBytes(uint32_t componentType) {
   if (componentType == TINYGLTF_COMPONENT_TYPE_BYTE) {
     return 1;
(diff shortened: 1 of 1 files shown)


syoyo:
The spec says emissiveFactor is `number[3]`, so I think we should report it as error, not warning.

https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-material

What is gltf-validator's result to emissiveFactor with `number[4]`?

nyalldawson:
How about if I add an optional "flags" argument to model loading, with a "permissive" flag which enables workarounds like this?

syoyo:
Adding a flag to control the parsing behavior would be a nice idea. For example...

`TinyGLTF::SetParseStrictness(enum)`

* permisive: allow some corrupted glTF.
* strict: current default
* validate: Do glTF validation during parsing(stricter) https://github.com/syoyo/tinygltf/issues/11

nyalldawson:
@syoyo how's that now?

syoyo:
> @syoyo how's that now?

Thanks! Let me give some time to review it.

syoyo:
Merged!

PR #446, opened 2023-08-24
Fix misleading error message

nyalldawson:
Avoids a confusing "Must have 4 bytes or more bytes, but got 4." error.


syoyo:
Awesome! Thanks! Merged!

PR #447, opened 2023-08-24
Handle the situation where the recorded component type does not match the required type for the actual number of stored points

nyalldawson:
This situation arises when decoding certain malformed files, most notably it's seen in glb tiles from Google Earth's 3d tileset.

It's a port of the workaround used by Cesium native here:

https://github.com/CesiumGS/cesium-native/blob/d9172461e2d53f19d050ebc39f24a4f4ef5b9224/CesiumGltfReader/src/decodeDraco.cpp#L101

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -4873,8 +4873,9 @@ static bool GetAttributeForAllPoints(uint32_t componentType, draco::Mesh *mesh,
 }
 
 static bool ParseDracoExtension(Primitive *primitive, Model *model,
-                                std::string *err,
-                                const Value &dracoExtensionValue) {
+                                std::string *err, std::string *warn,
+                                const Value &dracoExtensionValue,
+                                ParseStrictness strictness) {
   (void)err;
   auto bufferViewValue = dracoExtensionValue.Get("bufferView");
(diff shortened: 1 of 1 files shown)


syoyo:
I think it'd be better to provide user callback functions for Draco decoding/encoding for flexibility, not implement workaround in the builtin Draco decoding function.

PR of callback functions for Draco decoding/encoding is much appreciated!

nyalldawson:
@syoyo would it be acceptable to you to follow the current approach but only use this logic when the permissive flag from https://github.com/syoyo/tinygltf/pull/445 is set?

syoyo:
@nyalldawson Yes, could you please modify this PR so that the situation is only considered when strictness flags is PERMISSIVE, like this?:

```
if (strictness == PERMISSIVE) {
  do PR 447
}
```

nyalldawson:
@syoyo ok, that's done. I've also made it add a warning string when the component type is upgraded when in permissive mode.

syoyo:
Awesome! Thanks! Merged!

TODO: This is still much appreciated 🥺 > PR of callback functions for Draco decoding/encoding is much appreciated!

PR #449, opened 2023-09-02
Always use size_t for byte offsets

emimvi:
Hi!

I had a glft file with a single large buffer which i then used in many bufferViews/accessors.
This ended up silently creating an invalid file with negative byteOffsets due to a size_t->int cast overflowing in serialization.

This request fixes that serialization bug, and uses size_t for all byte offsets, instead of a mix depending on struct (bufferView, accessor, accessor::Sparse)

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -837,7 +837,7 @@ struct Accessor {
     int count;
     bool isSparse;
     struct {
-      int byteOffset;
+      size_t byteOffset;
       int bufferView;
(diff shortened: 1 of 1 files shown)


syoyo:
Good catch!

https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_accessor_byteoffset

The spec says byteOffset is ">=0", so we can use size_t(uint).
Let me give some time to review the PR

syoyo:
Merged! Thanks!

PR #450, opened 2023-09-02
Fix msvc build -- STRICT is a msvc macro name

nyalldawson:
Follow up https://github.com/syoyo/tinygltf/pull/445

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -196,8 +196,8 @@ typedef enum {
 } Type;
 
 typedef enum {
-  PERMISSIVE,
-  STRICT
+  Permissive,
+  Strict
 } ParseStrictness;
 
 static inline int32_t GetComponentSizeInBytes(uint32_t componentType) {
@@ -1562,7 +1562,7 @@ class TinyGLTF {
   size_t bin_size_ = 0;
   bool is_binary_ = false;
 
-  ParseStrictness strictness_ = ParseStrictness::STRICT;
+  ParseStrictness strictness_ = ParseStrictness::Strict;
 
   bool serialize_default_values_ = false;  ///< Serialize default values?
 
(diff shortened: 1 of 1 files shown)


syoyo:
MSVC CI build passes. https://github.com/syoyo/tinygltf/pull/445

What is the error message?

nyalldawson:
Sorry, it's actually the msys build which has issues with this name. See eg https://github.com/qgis/QGIS/actions/runs/6060705066/job/16445155328

syoyo:
I see. it'd be better to add msys2 + MinGW(ucrt) CI to tinygltf's Github Actions before merging this PR.

nyalldawson:
Ok, https://github.com/syoyo/tinygltf/pull/451 adds that. BUT it passes, and it turns out it's not part of the mingw build itself which is defining this macro. 

It must be coming from one of the dependencies in the project. (There's some hefty ones there, like Qt and GDAL which have hundreds of dependencies themselves). A quick code scan on github shows a number of projects defining this macro: https://github.com/search?q=%22%23Define+strict+%22&type=code&p=1

More details here: https://learn.microsoft.com/en-au/windows/win32/winprog/enabling-strict?redirectedfrom=MSDN

syoyo:
I see.

I think it'd be better to add suffix(or prefix) to enum identifier or only capitalize the first character to avoid potential conflict of any macro names. How do you think?


```
  typedef enum {
    STRICTNESS_PERMISSIVE,
    STRICTNESS_STRICT,
  } ParseStrictness;

  // or

  typedef enum {
    Permissive,
    Strict,
  } ParseStrictness;

```

nyalldawson:
@syoyo 
> I think it'd be better to add suffix(or prefix) to enum identifier or only capitalize the first character to avoid potential conflict of any macro names. How do you think?

Ok, done. I've gone with the capitalization change to avoid any risk.

syoyo:
Brilliant! Thanks! Merged! > I've gone with the capitalization change to avoid any risk.

PR #453, opened 2023-09-06
Fix build with draco

nyalldawson:
Sorry, I missed one in https://github.com/syoyo/tinygltf/pull/450 ...


syoyo:
👍

syoyo:
Ah, Draco build is not included CI so CI couldn't detect this issue.

TODO: Having Github Action with Draco build

PR #456, opened 2023-10-05
Allow BufferView indices to be unspecified.

haroonq:
Allow BufferView indices for element array buffers to be unspecified to support some extensions.

Note that this is similar to how invalid array buffers are handled in order to support, for example, sparse morph targets.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -6145,20 +6145,22 @@ bool TinyGLTF::LoadFromString(Model *model, std::string *err, std::string *warn,
           return false;
         }
 
-        auto bufferView =
+        const auto bufferView =
             model->accessors[size_t(primitive.indices)].bufferView;
-        if (bufferView < 0 || size_t(bufferView) >= model->bufferViews.size()) {
+        if (bufferView < 0) {
+          // skip, bufferView could be null(-1) for certain extensions
+        } else if (size_t(bufferView) >= model->bufferViews.size()) { 
           if (err) {
             (*err) += "accessor[" + std::to_string(primitive.indices) +
                       "] invalid bufferView";
           }
           return false;
+        } else {
+          model->bufferViews[size_t(bufferView)].target =
+              TINYGLTF_TARGET_ELEMENT_ARRAY_BUFFER;
+          // we could optionally check if accessors' bufferView type is Scalar, as
+          // it should be
         }
-
-        model->bufferViews[size_t(bufferView)].target =
-            TINYGLTF_TARGET_ELEMENT_ARRAY_BUFFER;
-        // we could optionally check if accessors' bufferView type is Scalar, as
-        // it should be
       }
 
       for (auto &attribute : primitive.attributes) {


syoyo:
Thanks!

> Allow BufferView indices for element array buffers to be unspecified to support some extensions

I need to confirm this fulfills glTF spec. Could you please point to glTF spec page https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html  describing about it?

Also, do you have minimal reproducible glTF scene?

syoyo:
FYI this PR breaks unit test

https://github.com/syoyo/tinygltf/actions/runs/6417703978/job/17440742346?pr=456

```
tester.cc:177: FAILED:
  REQUIRE_THAT( err, Catch::Contains("accessor[0] invalid bufferView") )
with expansion:
  "" contains: "accessor[0] invalid bufferView"
```

haroonq:
I believe that draco-compressed GLTFs will exhibit this issue.  For example, [glTF-Draco/Box.gltf](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/Box/glTF-Draco/Box.gltf)

If you compare that asset with the "normal" GLTF [glTF/Box.gltf](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/Box/glTF/Box.gltf), you'll see that it doesn't specify a `bufferView` value in any of its accessors.

I believe this behaviour is described in the [KHR_draco_mesh_compression](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md) specification.

I realize that tinygltf does support dracro decompression (if you define TINYGLTF_ENABLE_DRACO), but for my use-case, I have my own implementation for handling the extension that I would like to use.  However, because tinygltf treats an unspecified index bufferView as an error, it stops parsing the rest of the file, preventing me from decompressing my own buffers.

(Specifically, I would like all the decoded vertex data to be stored in a single, contiguous buffer, whereas tinygltf will decompress each attribute into its own separate buffer.)

Finally, it's also worth noting from the main spec that index buffers can also be [sparse buffers](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#sparse-accessors.).    All I've done is copy the same logic for the vertex buffers to apply to the index buffers as well (See commit: [e4132167222](https://github.com/syoyo/tinygltf/commit/e4132167222efec26ba012e7481dd067a6e23cb0).)

syoyo:
@haroonq Thank you for the info!

> https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-accessor

Confirmed `Accessor.bufferView` is not a required attribute.

So do not report an error when `Accessor.bufferView` is `-1`.

Unit test failure is caused by bufferView index out-of-range, so the logic will be

```
if (bufferView < 0) {
  // skip
} else if (bufferView < bufferViews.size()) {
  // ok
} else {
  // report error
}
```

haroonq:
Thank you!  I've updated the request with your suggestions.

syoyo:
Thanks! Merged!

Issue #459, opened 2023-11-22
Default constructed Material has wrong emissiveFactor


ptc-tgamper:
**Describe the issue**

When one creates a tinygltf::Material instance, all member variables are initialized to the proper defaults according to [spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-material), except the `emissiveFactor`. The latter should be {0.0f, 0.0f, 0.0f} by default, but it is empty instead.

This didn't cause any issues when writing the material because the empty emissiveFactor was simply skipped, behaving as it was the default. Parsing is not a problem either, as emissiveFactor is always explicitly set to {0.0f, 0.0f, 0.0f} if the gltf material does not hold an emissiveFactor. Still, it is simply confusing when instantiating a tinygltf::Material oneself.

**To Reproduce**

```
tinygltf::Material mat{};
if (mat.emissiveFactor.empty())
{
 std::cout << "WRONG << std::endl;
}
```

- OS: ALL
- Compiler, compiler version, compile options: ALL

**Expected behaviour**

tinygltf::Material default constructor shall initialize emissiveFactor to {0.0f, 0.0f, 0.0f}. 

**Additional context**

PR #460, opened 2023-11-22
Fix Default constructed Material has wrong emissiveFactor #459

ptc-tgamper:
tiny_gltf.h - properly initialise emissiveFactor;
tests/tester.cc - add test case

### DIFF
--- a/tests/tester.cc
+++ b/tests/tester.cc
@@ -494,25 +494,23 @@ TEST_CASE("image-uri-spaces", "[issue-236]") {
 }
 
 TEST_CASE("serialize-empty-material", "[issue-294]") {
-
   tinygltf::Model m;
-
-  tinygltf::Material mat;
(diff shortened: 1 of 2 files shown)


syoyo:
Awesome!

It'd be better to use member initialization instead of initialize members in the constructor. Its more C++11 or later friendly


```
    std::vector<double> emissiveFactor{{0.0, 0.0, 0.0}};  // length 3. default [0, 0, 0]
    std::string alphaMode{"OPAQUE"};               // default "OPAQUE"
    double alphaCutoff{0.5};             // default 0.5
    bool doubleSided{false};             // default false;

   Material() = default;
```

There are some ancient code in tinygltf which initialize member variables in the constructor to support C++03 compiler(and gcc-4.8), but tinygltf now drops supporting C++03.

syoyo:
Awesome! Thanks! Merged!

Issue #457, opened 2023-11-06
Empty nodes are wrongly serialized as null


ptc-tgamper:
**Describe the issue**

GLTF allows nodes to be empty, but not to be null. But tinygltf serializes empty nodes as a json::null object. The underlying reason is that the default constructor for detail::json (be it nlohmann or rapidjson) initializes the json object to a null object.

**To Reproduce**

- OS: all
- Compiler, compiler version, compile options: all
- Code to reproduce
```
    tinygltf::Model m{};
    m.nodes.push_back({});
    m.scenes.push_back({});
    m.scenes.front().nodes.push_back(0);
    tinygltf::TinyGLTF writer;
    writer.WriteGltfSceneToFile(&m, "emptynode.gltf", false, false, true, false);
```

The resulting gltf is invalid:
```
{
  "asset": {
    "version": "2.0"
  },
  "nodes": [
    null
  ],
  "scenes": [
    {
      "nodes": [
        0
      ]
    }
  ]
}
```

**Expected behaviour**

The empty nodes should be serialised as an empty json object, i.e. {} instead of null.

Additional context:

Validation error message:
```
Incorrect type. Expected "object".
Matches a schema that is not allowed.
Type mismatch. Property value null is not a 'object'.glTF Validator(TYPE_MISMATCH)
Node

A node in the node hierarchy. When the node contains `skin`, all `mesh.primitives` **MUST** contain `JOINTS_0` and `WEIGHTS_0` attributes. A node **MAY** have either a `matrix` or any combination of `translation`/`rotation`/`scale` (TRS) properties. TRS properties are converted to matrices and postmultiplied in the `T * R * S` order to compose the transformation matrix; first the scale is applied to the vertices, then the rotation, and then the translation. If none are provided, the transform is the identity. When a node is targeted for animation (referenced by an animation.channel.target), `matrix` **MUST NOT** be present.
```

syoyo:
Good catch!

This reminds me this issue: https://github.com/syoyo/tinygltf/pull/295

Probably we also need to apply this fix to other glTF objects: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-gltf

ptc-tgamper:
It seems an empty scene causes the same problem.

syoyo:
You can contribute a fix! PR is much appreciated.

ptc-tgamper:
Will do.

syoyo:
🙏

PR #462, opened 2023-11-23
Fix Empty nodes are wrongly serialized as null Issue #457

ptc-tgamper:
tiny_gltf.h - make sure to serialize null node as empty object;
tester.cc - add respective test case


syoyo:
Thanks! Merged!

TODO: Do the same for other glTF `scene` items.

Issue #458, opened 2023-11-11
Light reference in the node


Tugcga:
**Describe the issue**

When create GLTF file for the scene with lights, actual reference to the light does not assigned to the node. 

**To Reproduce**

Minimal example with wrong output:

```
tinygltf::Model model;
tinygltf::Scene scene;

// setup the light
tinygltf::Light light;
light.name = "light";
light.type = "point";
light.intensity = 0.75;
light.color = std::vector<double>{1.0, 0.8, 0.95};

model.lights.push_back(light);

// setup the node
tinygltf::Node node;
node.name = "light_node";
// assign light to the node
node.light = 0;
// add to the model
model.nodes.push_back(node);
// and to the scene
scene.nodes.push_back(0);

scene.name = "scene_Name";
model.scenes.push_back(scene);
model.defaultScene = 0;

tinygltf::TinyGLTF gltf;
gltf.WriteGltfSceneToFile(&model, "output.gltf");
```

As a result it produce the following ```output.gltf```:
```
{
  "asset": {
    "version": "2.0"
  },
  "extensions": {
    "KHR_lights_punctual": {
      "lights": [
        {
          "color": [
            1.0,
            0.8,
            0.95
          ],
          "intensity": 0.75,
          "name": "light",
          "type": "point"
        }
      ]
    }
  },
  "extensionsUsed": [
    "KHR_lights_punctual"
  ],
  "nodes": [
    {
      "extensions": {
        "KHR_lights_punctual": {}
      },
      "name": "light_node"
    }
  ],
  "scene": 0,
  "scenes": [
    {
      "name": "scene_Name",
      "nodes": [
        0
      ]
    }
  ]
}
```

**Expected behaviour**

The node with the name ```light_node``` should contains
```
"extensions": {
        "KHR_lights_punctual": {
          "light": 0
        }
      },
```

**Possible solution**

In the ```static void SerializeGltfNode(const Node &node, detail::json &o)``` function find the ```KHR_lights_punctual``` section in the json by using ```detail::FindMember(extensions, "KHR_lights_punctual", it);``` instead of ```detail::FindMember(o, "KHR_lights_punctual", it);```

syoyo:
Good find!

The spec says https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_lights_punctual/README.md#adding-light-instances-to-nodes

So we need to list light ids in `extensions.KHR_lights_punctual` of `node`.

[...]

PR #463, opened 2023-11-23
Fix Light reference in the node issue #458

ptc-tgamper:
The bug is actually a crasher, if one has asserts enabled.

tiny_gltf.h - access correct json object when serializing a light, this fixes an assert and causes us to serialze the light index properly; tester.cc - add respective testcase


syoyo:
Super awesome! Thanks!

Issue #464, opened 2023-11-23
Empty scenes are wrongly serialized as null


ptc-tgamper:
**Describe the issue**

Empty scenes are incorrectly serialized as json null. They should be serialized as empty json objects instead. 

**To Reproduce**

```
tinygltf::Model m;
m.scenes.push_back({});
tinygltf::TinyGLTF ctx;
std::stringstream os;
ctx.WriteGltfSceneToStream(&m, os, false, false);
```
The output is as follows:
`{"asset":{"version":"2.0"},"scenes":[null]}`

- OS: ALL
- Compiler, compiler version, compile options: ALL

**Expected behaviour**

Empty GLTF scenes shall be serialized as empty JSON objects

syoyo:
Yes, as commented https://github.com/syoyo/tinygltf/issues/457#issuecomment-1794801943, I think we should create empty objects for other glTF scene items(e.g. `meshes`) also.

FYI, It looks glTF spec totally prohibits `null` JSON object
(No `null` keyword in https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html)

ptc-tgamper:
Yeah, I am doing this in small, easy digestible pieces.

PR #465, opened 2023-11-23
Fix Empty scenes are wrongly serialized as null Issue #464

ptc-tgamper:
tinygltf.h - serialize empty scenes as empty json objects;
tester.cc 
- add respective test case
- bring test cases in correct order
- tag test case for light index with correct issue number and fix it to compare to deserialized model

### DIFF
--- a/tests/tester.cc
+++ b/tests/tester.cc
@@ -756,76 +756,6 @@ TEST_CASE("load-issue-416-model", "[issue-416]") {
   REQUIRE(true == ret);
 }
 
-TEST_CASE("serialize-light-index", "[issue-457]") {
-  tinygltf::Model model;
-  tinygltf::Scene scene;
-
-  // Create the light
-  tinygltf::Light light;
-  light.type = "point";
-  light.intensity = 0.75;
-  light.color = std::vector<double>{1.0, 0.8, 0.95};
-  // Add the light to the model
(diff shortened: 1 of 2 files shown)


tests/tester.cc
```diff
@@ -756,34 +756,83 @@ TEST_CASE("load-issue-416-model", "[issue-416]") {
   REQUIRE(true == ret);
 }
 
-TEST_CASE("serialize-light-index", "[issue-457]") {
```
ptc-tgamper, line 759: This had the wrong issue number

tests/tester.cc
```diff
@@ -794,11 +843,11 @@ TEST_CASE("serialize-light-index", "[issue-457]") {
     bool ok = ctx.LoadASCIIFromString(&m, nullptr, nullptr, os.str().c_str(), os.str().size(), "");
     REQUIRE(true == ok);
     // Check if the light was correctly serialized
-    REQUIRE(1 == model.lights.size());
-    CHECK(model.lights[0] == light);
+    REQUIRE(1 == m.lights.size());
```
ptc-tgamper, line 846: Make sure to check against the parsed model, not the one for serialization

syoyo:
Super cool! And good find! > issue number, light index tester

PR #467, opened 2023-12-04
Small security and overflow patch

rhiskey:
Fixed issues with printing via `sprintf` and `header_and_json` file size.

### DIFF
--- a/stb_image_write.h
+++ b/stb_image_write.h
@@ -773,7 +773,7 @@ static int stbi_write_hdr_core(stbi__write_context *s, int x, int y, int comp, f
 #ifdef __STDC_LIB_EXT1__
       len = sprintf_s(buffer, sizeof(buffer), "EXPOSURE=          1.0000000000000\n\n-Y %d +X %d\n", y, x);
 #else
-      len = sprintf(buffer, "EXPOSURE=          1.0000000000000\n\n-Y %d +X %d\n", y, x);
+      len = snprintf(buffer, sizeof(buffer), "EXPOSURE=          1.0000000000000\n\n-Y %d +X %d\n", y, x);
 #endif
       s->func(s->context, buffer, len);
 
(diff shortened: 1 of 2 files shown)


rhiskey:
Please review if this change takes place:

https://github.com/syoyo/tinygltf/pull/467/commits/1c6f6efafcf10f8a2981a1316253e16a5a2ed546

rhiskey:
Reverted back original limits and added `#undef max` (in case of Windows)

syoyo:
@rhiskey Thanks!

For annoying MSVC min/max macro issue, the best way is wrapping it with braces(it works also for non-windows platform)

```
(std::max)(a, b)

(std::numeric_limits<float>::max)()
```

rhiskey:
> @rhiskey Thanks!
> 
> For annoying MSVC min/max macro issue, the best way is wrapping it with braces(it works also for non-windows platform)
> 
> ```
> (std::max)(a, b)
> 
> (std::numeric_limits<float>::max)()
> ```

Thank you very much for taking the time and telling me about this method. Added a new commit with MSVC "anti-annoying fix".

syoyo:
For `sprintf_s`,  as CI test fails, it is not guaranteed  to be available when  `__STDC_LIB_EXT1__ ` is not defined.

https://en.cppreference.com/w/c/io/fprintf

> As with all bounds-checked functions, printf_s , fprintf_s, sprintf_s, and snprintf_s are only guaranteed to be available if __STDC_LIB_EXT1__ is defined by the implementation and if the user defines __STDC_WANT_LIB_EXT1__ to the integer constant 1 before including [<stdio.h>](https://en.cppreference.com/w/c/io).

So please use `snprintf`, which is available from C99

https://stackoverflow.com/questions/59948723/how-to-use-sprintf-s-for-linux-c

rhiskey:
I really appreciate your attention. Hope this small fix will be fine

syoyo:
@rhiskey Thanks! Merged!

Issue #177, opened 2019-07-11
Separate text parsing from binary parsing


UX3D-schmithuesen:
It shouldn't be necessary to load the image files and separate binary buffer when loading the glTF asset – most actions and inspections don't require that data and we only want to load it if and when it is needed.

syoyo:
Please describe the issue in detail. What is your problem?

Ybalrid:
I think I understand: 

@UX3D-schmithuesen wish there could be a way to either skip or defer image and binary buffer loading. 

Currently the behavior of the library is; when you call on of the TinyGLTF::LoadX() functions, 

 - `JSON` metadata is parsed, creating the arrays of nodes, scenes, meshes, accessors, and other kind of glTF objects
 - Binary buffers are loaded (eventually decoded from base64 text) in memory into `vector`s of bytes
 - Images are also loaded, uncompressed (via, e.g: stb_image) and also loaded into `vector`s of bytes.

I think Benjamin has a use-case where he wants to use only the result of the parsing of the JSON, and doesn't actually need to load to memory either the binary buffers or the textures. At least not from the get go.

That's actually an interesting idea.

What you could do at the moment is, for the images, you can stub their loading out by providing your own image loader callback. You can do that by providing a function pointer to `void TinyGLTF::SetImageLoader(LoadImageDataFunction func, void *user_data)` and doing `#define TINYGLTF_NO_STB_IMAGE`. If your callback doesn't really load the image data, (it just says to tinygltf that it's done) that could already give you a significant speed boost. In my experience, loading the images of a glTF assets is the slowest part of the loading via `tinygltf`

@syoyo What is you thought about adding a "deferred loading" feature to tinygltf. Something as simple as having `buffer` and  `image` carry a "loaded/unloaded` flag. Use could call a `loadData()` function to trigger loading. We could keep the current behavior intact, but maybe introduce a `TINYGLTF_DEFFERED_LOADING` compile time option that adds this feature. That could be useful. What do you think?

UX3D-schmithuesen:
That is correct, but maybe I just didn't pay enough attention – if I define `TINYGLTF_NO_FS`, I believe that the tool doesn't read any files which would be exactly what I want.

syoyo:
`TINYGLTF_NO_FS` this is not documented in README and rather still for developers. 

[...]

PR #470, opened 2024-02-02
Parallel image parsing for low latency use cases

bekorn:
Briefly: I developed a gltf parser for my needs but want to switch to a 3rd party solution. Knew tinygltf was an option. Did some testing and it was slow. Wanted to apply the multithreading strategy I used on my parser.

I know that sophisticated projects have their thread managers and can do a better job. The case I optimized for is small apps with the sole purpose of displaying a single model as fast as possible, such as Windows 3D Viewer or gltf-viewers. Or any other case where the latency is important and there are idle cores. I observed x4 or so speedup with gltf files with 40MB+ image files.

The initial implementation with minimal changes I did looked like a hack, so I had to make larger changes. While reading the code, I realized some repetition and made a generic function, `ParseArrayProperty`, that unified many use cases (see `LoadFromString`). This function also resizes the vector before any insertion, which prevents any unnecessary allocations.

There is an issue though, I don't get the `required` argument of `Parse*Property` functions. It is just a flag that prevents error logging, while the name suggests something that should affect the program flow. Like, if that property is absent, parsing should stop. `LoadFromString` has such a section that some members are checked and if any are absent, returns false. Also, error logging can be suppressed by passing a nullptr to err as it is already checked.

I implemented the `ParseArrayProperty` with a different behaviour, which worked for LoadFromString but failed for others. Hence the _draft_ pull request. Do you have any suggestions on what to do? Should I change the argument's name or behaviour?

Also, I measured the time it takes to copy all the pixels from stb_image buffer to a vector, which takes quite a big portion of the parsing process. I think capturing it with a std::unique_ptr can totally eliminate that.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -4213,6 +4213,8 @@ static bool ParseImage(Image *image, const int image_idx, std::string *err,
                        bool store_original_json_for_extras_and_extensions,
                        const std::string &basedir, const size_t max_file_size,
                        FsCallbacks *fs, const URICallbacks *uri_cb,
+                       std::vector<Buffer> const & buffers,
+                       std::vector<BufferView> const & bufferViews,
                        LoadImageDataFunction *LoadImageData = nullptr,
                        void *load_image_user_data = nullptr) {
   // A glTF image must either reference a bufferView or an image uri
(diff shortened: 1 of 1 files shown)


syoyo:
If the parsing speed is dominated by image decoding,  I think preferred approach would be using delayed decoding strategy.

https://github.com/syoyo/tinygltf/issues/177#issuecomment-995777019

This will keep code change small. PR of having delayed(threaded) decoding of images by default would be nice. 

For refactoring `Parse*Property`.  Please create another PR(or Issue) to discuss it.

PR #475, opened 2024-03-19
M_PI was not defined by <cmath>

jam3sward:
`<cmath>` does not define `M_PI` on WIN32. Whilst we could define `_USE_MATH_DEFINES` in the project settings, or in code, it seems better to simply define a constant. It's only used at this one point, so makes sense to do it locally.


syoyo:
Thanks! Merged!

PR #478, opened 2024-03-25
Remove Parameter class and respective dead code

ptc-tgamper:
tiny_gltf.h - remove Parameter class and associated code
loader_example.cc - likewise

Removing this to reduce the complexity of the parsing code. Should make work on my next patches easier.

### DIFF
--- a/loader_example.cc
+++ b/loader_example.cc
@@ -212,27 +212,6 @@ static std::string Indent(const int indent) {
   return s;
 }
 
-static std::string PrintParameterValue(const tinygltf::Parameter &param) {
-  if (!param.number_array.empty()) {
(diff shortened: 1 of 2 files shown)


syoyo:
Good catch.

But as commented like here:

https://github.com/syoyo/tinygltf/blob/release/tiny_gltf.h#L5345

Someone might still use ParameterMap in their app, so I'd like to postpone merging this PR until next major release(no ETA for next major release at the moment though).

ptc-tgamper:
Would draco compression support justify a new major release?

syoyo:
draco compression support  would result in a minor release(minior version bump)

PR #479, opened 2024-03-25
Added the pmode argument to _wopen to fix the access permission on MinGW

### LINKED ISSUE #477: WriteGltfSceneToFile (_wopen) might fail when compiling with MinGW (MSYS2)
**Describe the issue**

The function_wopen might fail to open an existing file created by the previous call to _wopen since the file is created as read-only by default. Passing  _S_IREAD | _S_IWRITE as the third argument solves the issue. 

**To Reproduce**
Call WriteGltfSceneToFile (WriteBinaryGltfFile) twice. 
- OS Windows 10
- Compiler g++ 13.2.0

**Expected behaviour**

Files should be created with the read and write permissions.

**Additional context**

[A related question on Stackoverflow](https://stackoverflow.com/questions/5516011/using-mingw-how-can-i-create-a-file-without-the-read-only-file-attribute)
[Microsoft Learn](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen)

> The pmode argument is required when _O_CREAT is specified

The0Dev:
Fixes #477

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -3151,7 +3151,7 @@ bool WriteWholeFile(std::string *err, const std::string &filepath,
 #ifdef _WIN32
 #if defined(__GLIBCXX__)  // mingw
   int file_descriptor = _wopen(UTF8ToWchar(filepath).c_str(),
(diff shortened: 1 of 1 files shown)


syoyo:
Thanks! Merged!

PR #480, opened 2024-03-25
Bug/msft lod extension used

ptc-tgamper:
At the moment the MSFT_lod extension is only registered with the nodes and materials that use it. But the model itself does not expose it as a used extension. Add MSFT_lod to the model's `extensionsUsed`, if one or more nodes/materials employ MSFT_lod.


syoyo:
Good find! Thanks for PR! Merged!

PR #481, opened 2024-03-25
Fix C4267 warnings in Visual Studio on WIN32

jam3sward:
This branch fixes C4267 warnings in Visual Studio 2022 on WIN32 when compiling with warning level 4.
It replaces int with size_t and removed the various casts. It appears that all the places that were using Get were already using it with unsigned values.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -337,12 +337,11 @@ class Value {
   T &Get();
 
   // Lookup value from an array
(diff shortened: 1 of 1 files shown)


syoyo:
Thanks! Merged!

I have no idea why I did use `int` for `Get`...

jam3sward:
Thanks, much appreciated!

PR #482, opened 2024-03-26
Fix C4018 warnings in MSVC on WIN32

jam3sward:
This fixes two C4018 (signed/unsigned mismatch) warnings in MSVC on WIN32 by using const auto, and removing some type casts.


jam3sward:
Ah, oops, it builds successfully in VS 2022 17.9.4, but the AppVeyor build fails. Will investigate.

jam3sward:
Fixed, and it seems happy now :)

syoyo:
Thanks! Merged!

PR #485, opened 2024-05-10
msvc 32bit: Fix C4244 warning

bwrsandman:
On 32 bit msvc compilers with warnings on, there are C4244 warnings about implicit casting from 'std::streamoff' to size_t and vector::size_type


syoyo:
Thanks! Merged!
(Sorry for the delay. I have missed this PR for some reason)

Issue #473, opened 2024-03-14
Images are not saved to disk properly


ptc-brainer:
**Describe the issue**

Writing a glTF that contains images referenced as uris without embedding images does not save the images to the disk next to the glTF model. 

**To Reproduce**

OS: MacOS
Compiler: Apple clang version 15.0.0

The following unit test fails with with the latest version of tinygltf:

```
TEST_CASE("write-image-issue") {

  tinygltf::Model model;
  tinygltf::TinyGLTF ctx;
  std::string err;
  std::string warn;

  REQUIRE(ctx.LoadASCIIFromFile(&model, &err, &warn, "../models/Cube/Cube.gltf"));

  REQUIRE(model.images.size() == 2);

  REQUIRE(model.images[0].uri == "Cube_BaseColor.png");
  REQUIRE(model.images[1].uri == "Cube_MetallicRoughness.png");

  REQUIRE_FALSE(model.images[0].image.empty());
  REQUIRE_FALSE(model.images[1].image.empty());

  REQUIRE(ctx.WriteGltfSceneToFile(&model, "Cube.gltf", false, true));

  for (const auto image : model.images)
  {
    std::fstream file(image.uri);
    CHECK(file.good());
  }
}
```

**Expected behaviour**

I would expect that the writer saves the images to the disk at the relative location described in the image uri.

**Additional context**

I traced this issue down to the `WriteImageData()` function that is passed to `UpdateImageObject()`. This function expects the `FsCallback` passed as the user data. However, the default `write_image_user_data` with is initialized with`nullptr`, causing the file to never be written to disk. An experimental change the initialize `write_image_user_data` with  a pointer to the default FsCallback (`&fs`) fixed this issue, but caused a different unit test failure.

syoyo:
Confirmed the issue.

Passing `FsCallback` to `write_image_user_data` is a confusing, so we are better to modify `WriteImageDataFunction` API to accept `FsCallbacks` callback pointer as an argument.

something like...

```
  typedef bool (*WriteImageDataFunction)(const std::string *basepath,
                                         const std::string *filename,
                                         const Image *image, bool embedImages,
                                         const FsCallback *fs_cb, 
                                         const URICallbacks *uri_cb,
                                         std::string *out_uri,
                                         void *user_pointer);
```

We may also need to add `overwrite existing image file` flag to the argument.

ptc-brainer:
Thanks for the quick investigation! Your proposal looks good to me!
[...]

PR #491, opened 2024-06-25
Fix images not being saved due to missing filesystem callbacks

ptc-tgamper:
tiny_gltf.h - explicitly pass filesystem callbacks to image related functions
tester.cc - add respective test case, fix image uri test case

### DIFF
--- a/tests/tester.cc
+++ b/tests/tester.cc
@@ -1060,11 +1060,10 @@ TEST_CASE("serialize-lods", "[lods]") {
 }
 
 TEST_CASE("write-image-issue", "[issue-473]") {
-
-  tinygltf::Model model;
-  tinygltf::TinyGLTF ctx;
   std::string err;
   std::string warn;
+  tinygltf::Model model;
+  tinygltf::TinyGLTF ctx;
   bool ok = ctx.LoadASCIIFromFile(&model, &err, &warn, "../models/Cube/Cube.gltf");
   REQUIRE(ok);
   REQUIRE(err.empty());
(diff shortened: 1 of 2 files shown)


tests/tester.cc
```diff
@@ -474,7 +474,7 @@ TEST_CASE("image-uri-spaces", "[issue-236]") {
     }
     REQUIRE(true == ret);
     REQUIRE(err.empty());
-    REQUIRE(!warn.empty());  // relative image path won't exist in tests/
+    REQUIRE(warn.empty());
```
ptc-tgamper, line 477: This test was relying on the buggy behaviour, i.e. a warning appearing because the image could not be loaded (because it wasn't saved in the first place).

tiny_gltf.h
```diff
@@ -1396,6 +1363,40 @@ bool GetFileSizeInBytes(size_t *filesize_out, std::string *err,
                         const std::string &filepath, void *);
 #endif
 
+///
```
ptc-tgamper, line 1364: Had to move this code block until after FsCallbacks was defined.

syoyo:
Thanks!

This PR solves the issue as you mentioned: https://github.com/syoyo/tinygltf/issues/473
And since it breaks FS API,  I need to be a bit carefully on merging this PR.

Just want to confirm: The difference is only adding FSCallback to FS Writer API? No other changes to FS API?

ptc-tgamper:
No other changes. Just directly passing the FSCallbacks through. Now the user pointer is actually the user pointer one would expect in all the callbacks.

tiny_gltf.h
```diff
@@ -2629,11 +2645,19 @@ bool LoadImageData(Image *image, const int image_idx, std::string *err,
   // the Image metadata to signal that this image uses 2 bytes (16bits) per
   // channel:
   if (stbi_is_16_bit_from_memory(bytes, size)) {
-    data = reinterpret_cast<unsigned char *>(
-        stbi_load_16_from_memory(bytes, size, &w, &h, &comp, req_comp));
-    if (data) {
-      bits = 16;
-      pixel_type = TINYGLTF_COMPONENT_TYPE_UNSIGNED_SHORT;
+    if (option.as_is) {
```
ptc-tgamper: This needs some rework before merging. But functionality should be there. Will clean up tomorrow.

ptc-tgamper:
@tdapper could you give this branch a try?

tiny_gltf.h
syoyo: Thanks!

ptc-tgamper: Done. Please review.

DanielLorenzLaubwerk:
I can confirm that the small test case does not crash anymore. I am going to test it in our original context now.

DanielLorenzLaubwerk:
It also works in our original context. 
Many thanks.

tiny_gltf.h
syoyo: Thanks. Let me give some time to review it

syoyo: In default `LoadImageData`, even as_is flag is set, it tries to decode image header(width, height, ...)
Can we make loading image header optional when `as_is` flag is set?
i.e., when failed to parse image header using STBI,  fill image buffer but set invalid size(default value) to other parameters(width, height, ...), report warning, and return true;

ptc-tgamper: I can do that.

ptc-tgamper: Done

tiny_gltf.h
```diff
@@ -2734,28 +2773,37 @@ bool WriteImageData(const std::string *basepath, const std::string *filename,
...
+  }
+
   if (ext == "png") {
-    if ((image->bits != 8) ||
-        (image->pixel_type != TINYGLTF_COMPONENT_TYPE_UNSIGNED_BYTE)) {
-      // Unsupported pixel format
-      return false;
-    }
+    if (!image->as_is) {
+      if ((image->bits != 8) ||
```
ptc-tgamper, line 2783: Perform these checks only if the image was decoded.

syoyo:
@ptc-tgamper Code is now OK(reviewed & CI test passes). I think its ready to merge. Do you have any further commits planned?

ptc-tgamper:
Not at the moment. Please merge, as this gets memory consumption down by an order of magnitude for our use case.

syoyo:
@ptc-tgamper Thanks! Merged!

syoyo:
Created new release FYI: https://github.com/syoyo/tinygltf/releases/tag/v2.9.0

PR #492, opened 2024-07-01
Make inverseBindMatrices optional

danwillm:
The spec mentions that `inverseBindMatrices` in skins are optional, but tinygltf produces an error when trying to load a model with none defined.

```
Each skin is defined by a REQUIRED joints property that lists the indices of nodes used as
joints to pose the skin and an OPTIONAL inverseBindMatrices property that points to
an accessor with inverse bind matrices data used to bring coordinates being skinned into 
the same space as each joint.
```


syoyo:
Good find! Confirmed https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-skin

> tinygltf produces an error when trying to load a model with none defined.

Could you possible to contribute a minimal glTF file which contains skin but no `inverseBindMatrices`? I'd like to add it to unit tests.

danwillm:
Just added a test for it. I can't upload the exact glb I'm trying to parse but I've made a mimimal example and uploaded it.

syoyo:
Thanks! Merged!

PR #493, opened 2024-07-04
Properly clear the model before loading

ptc-tgamper:
Not all, but only some, model fields were cleared before parsing.
As one can pass an existing model to the load functions, we should make sure to completely wipe the model before parsing.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -6108,16 +6108,8 @@ bool TinyGLTF::LoadFromString(Model *model, std::string *err, std::string *warn,
     }
   }
 
-  model->buffers.clear();
-  model->bufferViews.clear();
-  model->accessors.clear();
-  model->meshes.clear();
-  model->cameras.clear();
-  model->nodes.clear();
-  model->extensionsUsed.clear();
-  model->extensionsRequired.clear();
-  model->extensions.clear();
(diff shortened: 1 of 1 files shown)


syoyo:
I have no idea why I coded manually initialize each members there. 

I think it is better to use default constructor to initialize Model. How do you think?

```
(*model) = Model();
```

marco-langer:
`clear()` keeps the already allocated memory if the model was already used before. Default-constructing a new model releases all preallocated memory.

Maybe that was the original intention here?

ptc-tgamper:
I just followed the existing code and completed it. But I can simply default construct a new model, that makes the entire block into a one-liner.

ptc-tgamper:
Done

syoyo:
> `clear()` keeps the already allocated memory if the model was already used before. Default-constructing a new model releases all preallocated memory.
> 
> 
> 
> Maybe that was the original intention here?

Memory efficiency was not considered.

TinyGLTF was first written for C++03, and I think at that time I don't trust myself I put all initializations in the constructor.

Now, in C++11, we can initialize member variable in member declaration, so its safe to assume a default constructor initializes all members correctly.

syoyo:
Thanks! Merged!

PR #494, opened 2024-07-05
Add ability to disable image loading

ptc-tgamper:
Hi. This is a PR which allows to completely disable image loading. This is important if one has glTF models with gigabytes of external images, or a huge GLB where most of the data is images.

Image saving is still correctly handled, by looking for the source model's image files, and copying them over to the directory where the target glTF is written.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -1213,6 +1213,9 @@ class Model {
 
   bool operator==(const Model &) const;
 
+  // The base directory of the glTF model file
+  std::string baseDir;
+
   std::vector<Accessor> accessors;
   std::vector<Animation> animations;
   std::vector<Buffer> buffers;
@@ -1554,6 +1557,15 @@ class TinyGLTF {
   bool GetImagesAsIs() const { return images_as_is_; }
 
   ///
+  /// Specify whether image data is loaded into memory during parsing
+  ///
+  void SetLoadImages(bool onoff) {
+      load_images_ = onoff;
+  }
+
+  bool GetLoadImages() const { return load_images_; }
+
+  ///
   /// Set maximum allowed external file size in bytes.
   /// Default: 2GB
   /// Only effective for built-in ReadWholeFileFunction FS function.
@@ -1589,6 +1601,8 @@ class TinyGLTF {
                                           /// RGBA) for backward compatibility.
 
   bool images_as_is_ = false; /// Default false (decode/decompress images)
+
+  bool load_images_ = true; /// Default true (load images)
 
   size_t max_external_file_size_{
       size_t((std::numeric_limits<int32_t>::max)())};  // Default 2GB
@@ -1655,6 +1669,7 @@ class TinyGLTF {
 #include <sys/stat.h>  // for is_directory check
 
 #include <cstdio>
+#include <cstdlib>
 #include <fstream>
 #endif
 #include <sstream>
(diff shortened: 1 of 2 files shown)


tiny_gltf.h
```diff
@@ -1213,6 +1213,9 @@ class Model {
 
   bool operator==(const Model &) const;
 
+  // The base directory of the glTF model file
+  std::string baseDir;
```
ptc-tgamper, line 1217: This is needed to find the model's image files

Question is if operator== for Model should honor this or not? Compare only the contents of the model, or also where it was loaded from?

tiny_gltf.h
```diff
@@ -2228,8 +2243,57 @@ static std::string JoinPath(const std::string &path0,
...
+    // Apple platforms onyl have the outdated variant of the
+    // realpath() function.
+    char buffer[PATH_MAX] = {0};
+    // Note that realpath() fails for paths that do not exist.
+    // Still, that is sufficient for our use case.
+    char *result = realpath(path.c_str(), buffer);
+    return (result != nullptr) ? std::string(buffer) : "";
+#else
+    std::string result;
+    char *full = realpath(path.c_str(), nullptr);
```
ptc-tgamper, line 2271: Seems this is not working on MinGW, need to find a fix.

tiny_gltf.h
```diff
@@ -3279,23 +3347,26 @@ static std::string MimeToExt(const std::string &mimeType) {
...
-  // If image has uri, use it as a filename
-  if (image.uri.size()) {
-    std::string decoded_uri;
-    if (!uri_cb->decode(image.uri, &decoded_uri, uri_cb->user_data)) {
-      // A decode failure results in a failure to write the gltf.
-      return false;
+  std::string decoded_uri;
+  if (!image.uri.empty()) {
+    // If image has uri, use it as a filename if it is not a data uri
+    if (!IsDataURI(image.uri)) {
```
ptc-tgamper, line 3362: We need to handle data URI here, as we leave images with data URIs as is, instead of loading the data and clearing their URI.

tiny_gltf.h
```diff
@@ -3313,12 +3384,35 @@ static bool UpdateImageObject(const Image &image, std::string &baseDir,
...
+      Image imageToWrite = image;
+      // If the image references an external source, but holds no data,
+      // external image was disabled for this image, or the image failed
+      // to load during parsing. Try to load the image file from the
+      // external file, and if successful, mark the image "as is" for
+      // writing.
+      bool sourceIsTarget = false;
+      if (!decoded_uri.empty() && imageToWrite.image.empty()) {
+          std::string sourceFilePath;
+          if (LoadExternalFile(&imageToWrite.image, &sourceFilePath, nullptr, nullptr, decoded_uri, sourceBaseDir,
```
ptc-tgamper, line 3397: Load the source image, and if able to do so write it to target directory.

tiny_gltf.h
```diff
@@ -2522,6 +2586,10 @@ static bool LoadExternalFile(std::vector<unsigned char> *out, std::string *err,
     return false;
   }
 
+  if (foundFile) {
```
ptc-tgamper, line 2589: Return the filepath that was actually found

syoyo:
I recommend to use your own image load/save handler to control image loading for a while.

ptc-tgamper:
I tried that, but having my own image loader and writer is not sufficient to solve the problem. Loading works with my own image loader, but writing doesn't.  Images that do not hold any data are not passed to the image writer routine. Moreover, images that use data URIs are also not possible to handle, because their URIs are wiped outside of the image loader, so one loses their data. Only images that have their data in buffers are handled correctly.

I can try to come up with a minimal changeset to fix those deficiencies, so that I can skip image loading, if I use custom image loaders and writers.

syoyo:
You are better first file a github issue.

PR #496, opened 2024-07-09
Allow WriteImageDataFunction() callback to be called with empty images

ptc-tgamper:
Allow WriteImageDataFunction() callback to be called with empty images.

The motivation on our side is that we want to avoid to load any image data into memory at all, as we have to deal with models that use multiple Gigabytes (compressed size) of textures. Thus we are looking to employ a custom image loader and writer, where the first does not load/decode any image data, and the second is still called for those empty images.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -2767,6 +2767,12 @@ bool WriteImageData(const std::string *basepath, const std::string *filename,
                     const Image *image, bool embedImages,
                     const FsCallbacks* fs_cb, const URICallbacks *uri_cb,
                     std::string *out_uri, void *) {
+  // Early out on empty images, report the original uri if the image was not written.
+  if (image->image.empty()) {
+    *out_uri = *filename;
+    return true;
+  }
+
   const std::string ext = GetFilePathExtension(*filename);
 
   // Write image to temporary buffer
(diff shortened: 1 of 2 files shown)


tiny_gltf.h
```diff
@@ -4434,8 +4447,18 @@ static bool ParseImage(Image *image, const int image_idx, std::string *err,
...
-  return LoadImageData(image, image_idx, err, warn, 0, 0, &img.at(0),
-                       static_cast<int>(img.size()), load_image_user_data);
+
+  bool ok = LoadImageData(image, image_idx, err, warn, 0, 0, &img.at(0),
+                          static_cast<int>(img.size()), load_image_user_data);
+
+  // If the image is encoded as data URI, and the actual image data was
+  // not extracted, then we keep the original data URI that holds all
+  // the data.
+  if (IsDataURI(uri) && ok && image->image.empty()) {
```
ptc-tgamper: I did not want to keep the URI unconditionally, because with the default loader one would have the data in memory twice: in `image->image`, and in `image->uri`.

syoyo:
Please create PR for each issues(if issues are orthogonal)

tiny_gltf.h
```diff
@@ -2767,6 +2767,12 @@ bool WriteImageData(const std::string *basepath, const std::string *filename,
                     const Image *image, bool embedImages,
                     const FsCallbacks* fs_cb, const URICallbacks *uri_cb,
                     std::string *out_uri, void *) {
+  // Early out on empty images, report the original uri if the image was not written.
+  if (image->image.empty()) {
```
ptc-tgamper, line 2771: With this we keep the previous behaviour, although the callback is now invoked for empty images.

tiny_gltf.h
```diff
@@ -3309,11 +3315,12 @@ static bool UpdateImageObject(const Image &image, std::string &baseDir,
...
-  // If callback is set and image data exists, modify image data object. If
-  // image data does not exist, this is not considered a failure and the
-  // original uri should be maintained.
+  // If callback is set, modify image data object.
+  // Note that the callback is also invoked for images without data.
+  // The default callback implementation simply returns true for
+  // empty images and sets the out URI to filename.
   bool imageWritten = false;
-  if (WriteImageData != nullptr && !filename.empty() && !image.image.empty()) {
+  if (WriteImageData != nullptr && !filename.empty()) {
```
ptc-tgamper, line 3323: Invoke the callback for empty images

ptc-tgamper:
I tailored this branch now to the specific issue, nothing more. Please review.
I will open separate issues for the other changes.

syoyo:
You should describe the summary of what this PR is.

ptc-tgamper:
Is it ok to merge now, or am I still missing something? @syoyo ?

syoyo:
Good! Merged! The description become clear.

PR #498, opened 2024-07-31
fixes #497 Keep image data URI if no image data is loaded

### LINKED ISSUE #497: glTF images with data URI lose their URI during parsing
**Describe the issue**

During parsing  of glTF Images that store their data in a data URI, the latter is discarded. Only images that reference an external file keep their URI, see https://github.com/syoyo/tinygltf/blob/10b23b6af22c8692434690a06ffe84e3fc12f561/tiny_gltf.h#L4398
Therefore, when one uses a custom image loader that leaves `image.image` empty, the image data cannot be loaded later from the original data URI.

**To Reproduce**

- OS
- Compiler, compiler version, compile options
- Please attach minimal and reproducible .glTF file if you got an issue related to .glTF data

**Expected behaviour**

The image's URI should not be discarded if it is not loaded into `image.image`

**Screenshots**

**Additional context**
We are using a custom image loaderto avoid loading images into memory, since our models have Gigabytes of images. Naturally then the custom image writer needs to have access to the original image data. But the original image.uri has been lost during parsing.

ptc-tgamper:
Do not discard image data uri if uri still needed

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -3293,15 +3293,17 @@ static bool UpdateImageObject(const Image &image, std::string &baseDir,
                               void *user_data, std::string *out_uri) {
   std::string filename;
   std::string ext;
-  // If image has uri, use it as a filename
-  if (image.uri.size()) {
-    std::string decoded_uri;
-    if (!uri_cb->decode(image.uri, &decoded_uri, uri_cb->user_data)) {
-      // A decode failure results in a failure to write the gltf.
-      return false;
+  // If image has a uri, and it is not a data uri, use it as filename
+  if (!image.uri.empty()) {
+    if (!IsDataURI(image.uri)) {
+      std::string decoded_uri;
+      if (!uri_cb->decode(image.uri, &decoded_uri, uri_cb->user_data)) {
+        // A decode failure results in a failure to write the gltf.
+        return false;
+      }
+      filename = GetBaseFilename(decoded_uri);
+      ext = GetFilePathExtension(filename);
     }
-    filename = GetBaseFilename(decoded_uri);
-    ext = GetFilePathExtension(filename);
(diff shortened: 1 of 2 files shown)


tiny_gltf.h
```diff
@@ -3293,15 +3293,17 @@ static bool UpdateImageObject(const Image &image, std::string &baseDir,
...
   std::string ext;
-  // If image has uri, use it as a filename
-  if (image.uri.size()) {
-    std::string decoded_uri;
-    if (!uri_cb->decode(image.uri, &decoded_uri, uri_cb->user_data)) {
-      // A decode failure results in a failure to write the gltf.
-      return false;
+  // If image has a uri, and it is not a data uri, use it as filename
+  if (!image.uri.empty()) {
+    if (!IsDataURI(image.uri)) {
```
ptc-tgamper, line 3298: Now we can end up here with a data URI, so protect against that case

tiny_gltf.h
```diff
@@ -4442,8 +4444,17 @@ static bool ParseImage(Image *image, const int image_idx, std::string *err,
...
 
-  return LoadImageData(image, image_idx, err, warn, 0, 0, &img.at(0),
-                       static_cast<int>(img.size()), load_image_user_data);
+  bool ok = LoadImageData(image, image_idx, err, warn, 0, 0, &img.at(0),
+                          static_cast<int>(img.size()), load_image_user_data);
+
+  // If the image is encoded as data URI, and the actual image data was
+  // not extracted, then we keep the original data URI that holds all
+  // the data.
+  if (IsDataURI(uri) && ok && image->image.empty()) {
```
ptc-tgamper, line 4453: Keep the uri only if the data has not been loaded. Otherwise we have it in memory twice.

PR #501, opened 2024-08-08
Fix Animation extensions being loaded in place of Sampler extensions

msklywenn:
The title says it all.
Currently, Animation Sampler extensions are not loaded and Animation extensions are loaded to their member instead.


syoyo:
Good find!

Do you have minimal reproducible gltf file? I'd like to add it to regression tests.

msklywenn:
I do but, unfortunately, I’m not allowed to share it.

syoyo:
ok, the fix seems obvious so it'd be ready to merge. I wanted a test case for sure.

syoyo:
Merged!

PR #503, opened 2024-10-16
CMake: fix export install dir

nim65s:
Hi,

CMake exports are currently installed in `<prefix>/lib/cmake`, but according to https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure, this is not part of the standard search path.

Here is a simple fix.


syoyo:
Good! Thanks!

PR #504, opened 2024-10-16
CMake: allow opt-out of installing vendored headers

nim65s:
Hi,

I already have nlohmann/json & nothings/stb on my system, and would like to be able to install tinygltf without conflicts.

Here is a simple fix for that.


syoyo:
👍

PR #509, opened 2024-12-29
Added NO_FS definition for std::ofstream usage

NoirMorilec:
In tinygltf.h, there was usage of std::ofstream without defining NO_FS


syoyo:
Good find! Thanks! Merged!

PR #517, opened 2025-05-19
Removed TINYGLTF_USE_CPP14 option since it is unused

nepp95:
Hello,

Whilst browsing some compile options I wanted to use, I noticed this define is completely unused. Very trivial but thought I'd point it out. The define isn't used anywhere in the library to do something c++14 specific, whereas it did in the past (#201).

Kind regards,
Niels


syoyo:
Thank you!
Merged!

PR #532, opened 2026-03-02
Removing problematic using namespace in a header file

AnisB:
Remove `using namespace std` from public header

Removing the `using namespace std` directive from the public header to avoid polluting the global namespace of any consuming project. This is a breaking change for downstream consumers who may have been relying on implicit `std` symbol resolution — they will need to explicitly qualify any affected `std::` types or add their own `using` declarations in their own translation units.

### DIFF
--- a/tiny_gltf.h
+++ b/tiny_gltf.h
@@ -2451,9 +2451,8 @@ inline unsigned char from_hex(unsigned char ch) {
 }
 
 static const std::string urldecode(const std::string &str) {
-  using namespace std;
-  string result;
-  string::size_type i;
+  std::string result;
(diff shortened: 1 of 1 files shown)


syoyo:
Good find.

Although `using namespace std` should be better to remove it(its redundant), but its defined inside a non-inlined function. 
It should be safe to use `using` in a function and won't affect global namespace.

syoyo:
Merged!

AnisB:
Thanks for the reactivity!

In my case, the project I am using it has a namespace called string, so if tinygltf is inclued first, I am screwed (even if the global namespace is not polluted elsewhere).

